PR #7, opened 2021-07-19
Adds the remaining packages (For darwin) that are ready.

lawsonamzn:
Packages were tested on darwin/mac/built for ios

Note that these ONLY include packages that build from source code
by fetching from the official git repository / sourceforge / etc, or are permissive
and tiny enough to be committed as is.

Reviewers are welcome to look at the whole thing or just their own bits, but I need 2 approvals across the board at least..


lawsonamzn:
Note that the actual contents of the build scripts are direct ports from an internal repo with the only change being copyright headers.

The only 'major' change in this build is moving mikkelsen down into package-system, and updating the scripts.

Also note that the packages were actually tested on Darwin by building ech one.

package-system/mikkelsen/build_package_image.py
```diff
@@ -0,0 +1,88 @@
...
+        '--',
+        '-destination generic/platform=iOS'
+    ]),
+    'linux'     : ([
+        '-G', 'Ninja Multi-Config',
+        '-DCMAKE_C_COMPILER=clang-6.0', 
+        '-DCMAKE_CXX_COMPILER=clang++-6.0'
+    ], []),
+    'windows'   : ([
+        '-G', 'Visual Studio 15 2017',
```
lumberyard-employee-dm, line 43: We don't support VS2017 anymore, so we should change this to Visual Studio 16 at minimum.
Also for VS2019 the default architecture and host is x64, so the `-Ax64` and `Thost=x64` options aren't needed.

That being said I think we should data drive these options as much as possible.

Getting clang-6.0.0 on a modern Linux system that isn't Ubuntu 18.04 LTS is not easiest task.
Furthermore if the goal is to build the package on the user's system, then hard-coding clang-6.0 isn't really ideal when newer versions of clang are what available in the package system of the specific linux OS(i.e RedHat and ArchLinux comes with clang-12.0.0)

Being able to adjust the settings via a configuration file instead of a python script could increase the number of users that could build update the package one day.

lumberyard-employee-dm (approved):
Approving for now, since we need to get this in.

amznestebanpapp (approved):

moraaar (approved):
I've reviewed PhysX, NvCloth, poly2tri and v-hacd.

package-system/mikkelsen/build_package_image.py
lawsonamzn: OK.   I'll file an issue to do this.  My current goal is to transplant the 3p libraries from the internal repos to somewhere useful from O3DE without incurring a whole lot of change during the process.

package-system/glad-multiplatform/Findglad.cmake
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+set(MY_NAME "glad")
```
jiaweig-amzn, line 9: Glad in 3rd party is used by legacy only. Previously when moving 3rd party, legacy wasn't fully deprecated. But now, I think we don't need this anymore.
Atom has its own Glad, so not having this one won't affect Atom.

PR #11, opened 2021-07-23
Enable using local git repo

rgba16f:
Add command line parameters to the pull_and_build_from_git script to enable passing a local git repo and skipping all git commands

### DIFF
--- a/Scripts/extras/pull_and_build_from_git.py
+++ b/Scripts/extras/pull_and_build_from_git.py
@@ -52,8 +52,8 @@ The following keys can exist at the root level or the target-platform level:
                           to restrict building to a specific configuration rather than building all configurations 
                           (provided by the default value: ['Debug', 'Release'])
 * patch_file            : (optional) Option patch file to apply to the synced source before performing a build
-* local_repo            : (optional) Option to provide a path to the project source rather than getting it from github
-* git_skip              : (optional) Option to skip all git commands, requires local_repo
+* source_path            : (optional) Option to provide a path to the project source rather than getting it from github
+* git_skip              : (optional) Option to skip all git commands, requires source_path
 
 
 The following keys can only exist at the target platform level as they describe the specifics for that platform
@@ -858,11 +858,11 @@ if __name__ == '__main__':
                             action="store_true")
         parser.add_argument('--build-path',
                             help="Path to build the repository in. Defaults to {base_path}/temp.")
-        parser.add_argument('--local-repo',
-                            help='Path to an existing git repo',
+        parser.add_argument('--source-path',
+                            help='Path to a folder. Can be used to specify the git sync folder or provide an existing folder with source for the library.',
                             default=None)
         parser.add_argument('--git-skip',
-                            help='skips all git commands, requires local-repo to be provided',
+                            help='skips all git commands, requires source-path to be provided',
                             default=False)
 
         parsed_args = parser.parse_args(sys.argv[1:])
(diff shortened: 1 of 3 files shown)


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -751,12 +758,21 @@ def prepare_build(platform_name, base_folder, build_folder, package_root_folder,
     :param toolchain_file:      Option toolchain file to use for specific target platforms
     :param build_config_file:   The build config file to open from the base_folder
     :param clean:               Option to clean any existing build folder before proceeding
+    :param src_folder:          Option to manually specify the src folder
+    :param skip_git:            Option to skip all git commands, requires src_folder be supplied
```
spham-amzn, line 762: I think skip_git is implicitly True if src_folder is supplied, right? We may not need this

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -840,6 +858,12 @@ def prepare_build(platform_name, base_folder, build_folder, package_root_folder,
                             action="store_true")
         parser.add_argument('--build-path',
                             help="Path to build the repository in. Defaults to {base_path}/temp.")
+        parser.add_argument('--git-path',
+                            help='Path to an existing git repo',
+                            default=None)
+        parser.add_argument('--git-skip',
```
spham-amzn, line 864: See comment about this above. And git-path is ambigous, it should be '--local-repo'

lawsonamzn:
Ah, I see, this would be for local testing, I assume?

rgba16f:
> Ah, I see, this would be for local testing, I assume?

Yes, I was using this while I was iterating on the DirectXShaderCompiler code.

Scripts/extras/pull_and_build_from_git.py
rgba16f: I didn't want to assume that supplying the code folder meant that no git commands should be run. It could just be that they want the repo cloned to a specific folder.

Scripts/extras/pull_and_build_from_git.py
rgba16f: will do.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -52,6 +52,8 @@
                           to restrict building to a specific configuration rather than building all configurations 
                           (provided by the default value: ['Debug', 'Release'])
 * patch_file            : (optional) Option patch file to apply to the synced source before performing a build
+* local_repo            : (optional) Option to provide a path to the project source rather than getting it from github
+* git_skip              : (optional) Option to skip all git commands, requires local_repo
```
lawsonamzn, line 56: sorry to ask for another change, but could you change this name to something like "source_directory" instead of "local_repo"?  And then remove git_skip because "source_directory" would imply it?

The reason for this is that a local repo is actually a git concept.
In fact, if you instead of doing this change this: instead

On line 337, in clone_to_local function, after the line 
```
relative_src_dir = self.src_folder.name
```
add:
```
relative_src_dir = self.src_folder.name

if not self.package_info.git_url.startswith("http") and not self.package_info.git_url.startswith("git"):
            self.package_info.git_url = str(pathlib.Path(self.package_info.git_url).absolute())
```

Then whats going to happen is git will happily take absolute or relative paths to folders containing a git repository and treat it as local automatically.

eg, if you have a local git repository in d:\myprojects\blah (you have called git init, git add . , git commit -m initial)

if you then do a 'git clone "d:/myprojects/blah" it will actualy automatically understand it doesn't have to go to the internet.  

Perhaps thats a better way of working then anyway since it also allows you do stil use this script to apply patches?

rgba16f: I disagree that using local_repo or source_folder always means I want to skip the git commands.

As an example I will be working out the 3p instructions for FreeType soon, and I will use the pull_and_build script to sync down the repo, but I want to specify where to put it. In that case I was to specify source_folder but not skip the git commands.

lawsonamzn (approved):

spham-amzn (approved):

PR #13, opened 2021-08-03
Adds a build-from-source script to zlib for all platforms

lawsonamzn:
Android, Windows, IOS, Mac and LInux covered.
Built each locally.  Built on an internal jenkins too.  All green.


package-system/zlib/Findzlib.cmake.template
```diff
@@ -0,0 +1,44 @@
...
+# this file actually ingests the library and defines targets.
+set(TARGET_WITH_NAMESPACE "3rdParty::zlib")
+if (TARGET $${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+
+set(zlib_INCLUDE_DIR $${CMAKE_CURRENT_LIST_DIR}/zlib/include)
+set(zlib_LIBS_DIR $${CMAKE_CURRENT_LIST_DIR}/zlib/lib)
+
+if ($${PAL_PLATFORM_NAME} STREQUAL "Linux")
```
amznestebanpapp, line 18: are you using this to escape it for when it is generated? a better way is to wrap the variables you want to replace with `@` and then generating with `@ONLY`: https://cmake.org/cmake/help/latest/command/string.html#configure 
That way is easier to distinguish the variables that will be expanded vs the ones that will stay

amznestebanpapp (approved):

package-system/zlib/Findzlib.cmake.template
spham-amzn: These template files are processed by Python, not cmake, that is why they are escaped like this.

spham-amzn (approved):

package-system/zlib/Findzlib.cmake.template
amznestebanpapp: thanks for the clarification

lawsonamzn: I'd like to add a feature to pull_and_build_from_git that lets you just use a cmake find file as is.  I'm not sure it actually supports just not using templates.

spham-amzn: Sounds good to me

PR #14, opened 2021-08-09
3p Package for cityhash

AMZN-tpeng:
Signed-off-by: Peng <tonypeng@amazon.com>


lawsonamzn (approved):
We didn't make any changes to cityhash, did we?

moudgils (approved):

PR #16, opened 2021-08-16
Add ilmbase (OpenEXR) build script.

hershey5045:
# Testing:
* Tested locally on windows machine. Building ilmbase from source works. Using the artifacts to build o3de works
* Tested on a linux machine. Building ilmbase from source works


package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+               "-T",
+               "host=x64",
+               "-DOPENEXR_BUILD_PYTHON_LIBS=OFF",
+               "-DOPENEXR_BUILD_TESTS=OFF",
+               "-DOPENEXR_BUILD_UTILS=OFF",
+               "-DOPENEXR_BUILD_SHARED=OFF",
+               "-DOPENEXR_BUILD_STATIC=ON",
+               "-DCMAKE_INSTALL_PREFIX=\"../../ilmbase-windows/ilmbase\"",
+               "-DCMAKE_BUILD_TYPE=Debug",
+               "-DZLIB_ROOT=../../../zlib-windows/zlib"
```
spham-amzn, line 28: This is a situation that we may need to address in the script. We may need to either use a custom script instead of the using the cmake arguments here, so that we would:

1. Sync OpenEXR (this can still be done by the script)
2. Have the custom script do a git clone for zlib based on the appropriate tags, etc
3. Build zlib for that platform first, then apply the zlib argument here
I'm not sure we want to do this long term (relying on the README), but I guess its acceptable for now

package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+               "Xcode",
+               "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.13",
+               "-DOPENEXR_BUILD_PYTHON_LIBS=OFF",
+               "-DOPENEXR_BUILD_TESTS=OFF",
+               "-DOPENEXR_BUILD_UTILS=OFF",
+               "-DOPENEXR_BUILD_SHARED=OFF",
+               "-DOPENEXR_BUILD_STATIC=ON",
+               "-DCMAKE_INSTALL_PREFIX=\"../../ilmbase-mac/ilmbase\"",
+               "-DCMAKE_BUILD_TYPE=Release",
+               "-DZLIB_ROOT=../../../zlib-windows/zlib"
```
spham-amzn, line 78: don't you mean zlib-mac ?

spham-amzn: same here

package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+               "-G",
+               "Ninja",
+               "-DOPENEXR_BUILD_PYTHON_LIBS=OFF",
+               "-DOPENEXR_BUILD_TESTS=OFF",
+               "-DOPENEXR_BUILD_UTILS=OFF",
+               "-DOPENEXR_BUILD_SHARED=OFF",
+               "-DOPENEXR_BUILD_STATIC=ON",
+               "-DCMAKE_INSTALL_PREFIX=\"../../ilmbase-linux/ilmbase\"",
+               "-DCMAKE_BUILD_TYPE=Debug",
+               "-DZLIB_ROOT=../../../zlib-windows/zlib"
```
spham-amzn, line 99: don't you mean zlib-linux?

package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+               "-G",
+               "Ninja",
+               "-DOPENEXR_BUILD_PYTHON_LIBS=OFF",
+               "-DOPENEXR_BUILD_TESTS=OFF",
+               "-DOPENEXR_BUILD_UTILS=OFF",
+               "-DOPENEXR_BUILD_SHARED=OFF",
+               "-DOPENEXR_BUILD_STATIC=ON",
+               "-DCMAKE_INSTALL_PREFIX=\"../../ilmbase-linux/ilmbase\"",
+               "-DCMAKE_BUILD_TYPE=Release",
+               "-DZLIB_ROOT=../../../zlib-windows/zlib"
```
spham-amzn, line 111: same here

package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+            ],
+            "cmake_build_args":[
+               "-j"
```
spham-amzn, line 114: You can remove this

package-system/OpenEXR/build_config.json
```diff
@@ -0,0 +1,119 @@
...
+      "Windows":{
+         "Windows":{
+            "cmake_find_template":"Findilmbase.cmake.Windows",
+            "cmake_generate_args_debug":[
+               "-G",
+               "\"Visual Studio 16 2019\"",
+               "-A",
+               "x64",
+               "-T",
+               "host=x64",
```
lumberyard-employee-dm, line 20: The Architecture and toolset argument isn't needed when building with VS2019 and above

package-system/OpenEXR/build_config.json
hershey5045: Done!

package-system/OpenEXR/build_config.json
hershey5045: Removed.

package-system/OpenEXR/build_config.json
hershey5045: Will do, removed now.

spham-amzn (approved):

package-system/OpenEXR/build_config.json
lawsonamzn: I'd like to fix this but I don't think its in the scope of adding ILMBase to fix it.

Ideally we'd be able to declare dependencies on other packages here and it fetches and unpacks it using the normal 3p system.  It does not need to build the other package, right?  We already have auto build scripts for zlib too, which means other option is for it to build the package (using the command in the host file) and use its "output folder" (also listed in the host file) as the source of it.

lawsonamzn (approved):
I don't think we need to change anything in the scope of this OpenEXR but I'd love to actually improve the core scripts to handle dependencies better as a seperate PR.

package-system/OpenEXR/build_config.json
spham-amzn: That approach sounds good, but dependencies would need to be declared in main build scripts and potentially can be declared in the 'package_build_list*.json' files.

PR #17, opened 2021-08-17
Update to pull_and_build_from_git.py

spham-amzn:
- Add a new field 'cmake_find_source' for the build_config.json to accept a straight source file that will not be processed
- Keep existing 'cmake_find_template'
- Update DirectXShaderCompiler to use new field since it doesnt need templating
- Minor fix to DirectXShaderCompiler scripts

Signed-off-by: Steve Pham <spham@amazon.com>


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -325,11 +338,15 @@ def __init__(self, package_info, platform_config, base_folder, build_folder, pac
...
         self.cmake_find_template = cmake_find_template
+        self.cmake_find_source = cmake_find_source
         self.build_configs = platform_config.get('build_configs', ['Debug', 'Release'])
         self.prebuilt_source = prebuilt_source
         self.prebuilt_args = prebuilt_args
         self.skip_git = skip_git
 
+        if self.cmake_find_template is not None and self.cmake_find_source is not None:
+            raise BuildError("Bad build config file. 'cmake_find_template' and 'cmake_find_template' cannot both be set in the configuration.")            
+
```
rgba16f, line 349: This check is done in 2 places, PackageInfo::__init__ and BuildInfo::__init__, do we need both?

Additionally PackageInfo checks for both None, do we also need that check here?

spham-amzn: Yeah, I think PackageInfo::init is all we need. I'll remove the other checks

rgba16f (approved):

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -171,11 +174,18 @@ def _get_value(value_key, required=True, default=None):
...
+        self.cmake_find_template = _get_value("cmake_find_template", required=False)
+        self.cmake_find_source = _get_value("cmake_find_source", required=False)
         self.cmake_find_target = _get_value("cmake_find_target")
         self.cmake_find_template_custom_indent = _get_value("cmake_find_template_custom_indent", default=1)
         self.additional_src_files = _get_value("additional_src_files", required=False)
 
+        if self.cmake_find_template is not None and self.cmake_find_source is not None:
+            raise BuildError("Bad build config file. 'cmake_find_template' and 'cmake_find_template' cannot both be set in the configuration.")            
+        if self.cmake_find_template is None and self.cmake_find_source is None:
+            raise BuildError("Bad build config file. 'cmake_find_template' or 'cmake_find_template' must be set in the configuration.")
```
lumberyard-employee-dm, line 186: This can be simplified to be `if not self.cmake_find_template and not self.cmake_find_source`
```suggestion
        if self.cmake_find_template and self.cmake_find_source:
            raise BuildError("Bad build config file. 'cmake_find_template' and 'cmake_find_template' cannot both be set in the configuration.")            
        if not self.cmake_find_template and not self.cmake_find_source:
            raise BuildError("Bad build config file. 'cmake_find_template' or 'cmake_find_template' must be set in the configuration.")
```

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -803,15 +828,31 @@ def prepare_build(platform_name, base_folder, build_folder, package_root_folder,
...
+        if not cmake_find_template_path.is_file():
+            raise BuildError("Invalid 'cmake_find_template' entry in build config")
+
+    elif package_info.cmake_find_source is not None:
+
+        # Validate the cmake find source
+        if os.path.isabs(package_info.cmake_find_source):
+            cmake_find_source_path = pathlib.Path(package_info.cmake_find_source)
+        else:
+            cmake_find_source_path = base_folder_path / package_info.cmake_find_source
```
lumberyard-employee-dm, line 850: There is no reason to perform an isabs check, the pathlib library will replace previous components if it finds an absolute path component when appended paths: https://docs.python.org/3/library/os.path.html#os.path.join
> If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.

Therefore this can be changed to just be
```suggestion
        # Validate the cmake find source
        cmake_find_source_path = base_folder_path / package_info.cmake_find_source
```

lumberyard-employee-dm (approved):

Scripts/extras/pull_and_build_from_git.py
rgba16f: The BuildError strings both say cmake_find_template twice, Both need adjusting so the 2nd name is cmake_find_source.

Scripts/extras/pull_and_build_from_git.py
spham-amzn: Actually, we shouldn't even allow absolute paths here. I'll add an abs check here anyways as an assert

PR #19, opened 2021-08-18
Move 3p AWSNativeSDK windows to github

onecent1101:
## Details
Still working on other platforms: linux, mac, ios and andoid
2. Add custom build and install script for windows

Note: There is no update on AWSNativeSDK content, so this PR doesn't bump up version. This PR focuses on migration from perforce prebuild instruction to github build script (in order to narrow down the scope and reduce impact as much as possible). Build script has been verified locally, and it can be verified again once we need to do AWSNativeSDK major version update.

## Testing
Tested locally

Signed-off-by: Liu <liug@amazon.com>

### DIFF
--- a/package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
+++ b/package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
@@ -14,15 +14,27 @@ call::GenerateCommonCMakeArgs CMAKE_COMMON_ARGS
 
 REM Debug Shared
 call:ConfigureAndBuild Debug Shared
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
+)
 
 REM Debug Static
 call:ConfigureAndBuild Debug Static
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
+)
 
 REM Release Shared
 call:ConfigureAndBuild Release Shared
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
+)
 
 REM Release Static
 call:ConfigureAndBuild Release Static
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
+)
 
 ECHO "Custom Build for AWSNativeSDK finished successfully"
 exit /b 0
--- a/package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
+++ b/package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
@@ -9,9 +9,6 @@ REM
 SET SRC_PATH=temp\src
 SET BLD_PATH=temp\build
 
-SET CMAKE_COMMON_ARGS=""
-call::GenerateCommonCMakeArgs CMAKE_COMMON_ARGS
-
 REM Debug Shared
 call:ConfigureAndBuild Debug Shared
 IF %ERRORLEVEL% NEQ 0 (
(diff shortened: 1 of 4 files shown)


package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
```diff
@@ -0,0 +1,54 @@
...
+SET LIB_TYPE=%~2
+SET BUILD_SHARED=OFF
+IF %LIB_TYPE% EQU Shared (
+    SET BUILD_SHARED=ON
+)
+ECHO "CMake Configure %BUILD_TYPE% %LIB_TYPE% with %CMAKE_COMMON_ARGS%"
+call cmake -S %SRC_PATH% -B %BLD_PATH%\%BUILD_TYPE%_%LIB_TYPE% %CMAKE_COMMON_ARGS% -DBUILD_SHARED_LIBS=%BUILD_SHARED% -DCMAKE_BUILD_TYPE="%BUILD_TYPE%" -DCMAKE_INSTALL_BINDIR="bin/%BUILD_TYPE%" -DCMAKE_INSTALL_LIBDIR="lib/%BUILD_TYPE%"
+IF %ERRORLEVEL% NEQ 0 (
+    ECHO "CMake Configure %BUILD_TYPE% %LIB_TYPE% failed"
+    exit /b 1
```
junbo75, line 41: Note that this line will only exit the current ConfigureAndBuild function but not the entire process. I think you may want to terminate the execution of this batch when error happens at any time. Same for the exit call inside other functions.

lawsonamzn (commented):
do you actually intend to commit this?  I
t needs updates to the host file in the root of the repo, to invoke pull_and_build_from_Git.py 
Or is this just a draft?

package-system/AWSNativeSDK/build_AWSNativeSDK_windows.cmd
onecent1101: sounds good, will update

onecent1101:
> 
> 
> do you actually intend to commit this? I
> t needs updates to the host file in the root of the repo, to invoke pull_and_build_from_Git.py
> Or is this just a draft?

do you mean for other platform? for windows it works fine, host file does invoke pull_and_build_from_git.py, and after git clone the sdk source, it will use my custom build and install script instead.

lawsonamzn (approved):
This looks good for now.  NOte that this is a special case, we are not actually planning to update the packages already uploaded since this contains no actual changes for them.

onecent1101:
> This looks good for now. NOte that this is a special case, we are not actually planning to update the packages already uploaded since this contains no actual changes for them.

Updated PR details with this information.

junbo75 (approved):

spham-amzn:
Can this PR wait for the other platforms?

onecent1101:
> Can this PR wait for the other platforms?

As I am doing the migration one platform at a time, so it will be easier to send and merge PR one platform at a time, which also keeps PR scope small to review. Do you have any concern for merging it right now?

spham-amzn:
> 
> 
> > Can this PR wait for the other platforms?
> 
> As I am doing the migration one platform at a time, so it will be easier to send and merge PR one platform at a time, which also keeps PR scope small to review. Do you have any concern for merging it right now?

I dont see the need to publish it until its complete, unless there is a need to actually update the windows version

package-system/AWSNativeSDK/build_config.json
```diff
@@ -0,0 +1,25 @@
...
+   "package_url":"https://github.com/aws/aws-sdk-cpp",
+   "package_license":"Apache-2.0",
+   "package_license_file":"LICENSE",
+   "cmake_find_target":"FindAWSNativeSDK.cmake",
+   "cmake_find_template_custom_indent": 3,
+   "build_configs": ["Debug", "Release"],
+   "Platforms":{
+      "Windows":{
+         "Windows":{
+            "cmake_find_template":"FindAWSNativeSDK.cmake.Windows.template",
```
spham-amzn, line 15: not an issue with this PR because it was just updated yesterday, but the pull_and_build_from_git script now supports 'cmake_find_source' instead of 'cmake_find_template', which will not use the file as a template if you dont need any templatized parameters. That will make it so you dont have to escape the '$' with '$$' in the file.

spham-amzn (approved):

package-system/AWSNativeSDK/build_config.json
onecent1101: sounds good, will update in next PR

PR #22, opened 2021-08-19
Added RapidJSON 1.1.0

santorac:
Added RapidJSON 1.1.0


package_build_list_host_darwin.json
```diff
@@ -67,6 +67,7 @@
         "mikkelsen-1.0.0.4-mac": "package-system/mikkelsen-mac",
         "mikkelsen-1.0.0.4-ios": "package-system/mikkelsen-ios",
         "poly2tri-7f0487a-rev1-mac": "package-system/poly2tri-mac",
+		"RapidJSON-1.1.0-multiplatform": "package-system/RapidJSON-multiplatform",
```
lumberyard-employee-dm, line 70: The alignment is off in this file

package_build_list_host_linux.json
```diff
@@ -47,6 +47,7 @@
         "NvCloth-v1.1.6-4-gd243404-pr58-rev1-linux": "package-system/NvCloth-linux",
         "mikkelsen-1.0.0.4-linux": "package-system/mikkelsen-linux",
         "poly2tri-7f0487a-rev1-linux": "package-system/poly2tri-linux",
+		"RapidJSON-1.1.0-multiplatform": "package-system/RapidJSON-multiplatform",
```
lumberyard-employee-dm, line 50: repeat: alignment issues with specifying this 3rdParty library

package_build_list_host_windows.json
```diff
@@ -83,6 +83,7 @@
     "mikkelsen-1.0.0.4-android": "package-system/mikkelsen-android",
     "OpenMesh-8.1-rsv1-windows": "package-system/OpenMesh-windows",
     "pyside2-qt-5.15.1-rev2-windows": "package-system/pyside2-windows",
+	"RapidJSON-1.1.0-multiplatform": "package-system/RapidJSON-multiplatform",
```
lumberyard-employee-dm, line 86: repeat: alignment issues in this file with specifying rapidjson

lumberyard-employee-dm:
Since rapidjson 1.1.0 is a vanilla library that doesn't need to build any source code, we could use the [pull_and_build_from_get.py](https://github.com/o3de/3p-package-source/blob/main/Scripts/extras/pull_and_build_from_git.py) to pull the source code from the tencent [rapidjson repo](https://github.com/Tencent/rapidjson/tree/v1.1.0).

@spham-amzn Do you know if the pull and build script can skip the "building" part?

package_build_list_host_darwin.json
santorac: Oh, I was working on my laptop for the first time in forever, which I guess hasn't been set up properly. Thanks.

spham-amzn:
> 
> 
> Since rapidjson 1.1.0 is a vanilla library that doesn't need to build any source code, we could use the [pull_and_build_from_get.py](https://github.com/o3de/3p-package-source/blob/main/Scripts/extras/pull_and_build_from_git.py) to pull the source code from the tencent [rapidjson repo](https://github.com/Tencent/rapidjson/tree/v1.1.0).
> 
> @spham-amzn Do you know if the pull and build script can skip the "building" part?

Yeah, I thought the requirement for these scripts was to not include source. build_and_pull_from_git can use custom build and install scripts, so the build scripts can be empty, and the install script can assemble the package folders before they are tar'd up

spham-amzn:
> 
> 
> > Since rapidjson 1.1.0 is a vanilla library that doesn't need to build any source code, we could use the [pull_and_build_from_get.py](https://github.com/o3de/3p-package-source/blob/main/Scripts/extras/pull_and_build_from_git.py) to pull the source code from the tencent [rapidjson repo](https://github.com/Tencent/rapidjson/tree/v1.1.0).
> > @spham-amzn Do you know if the pull and build script can skip the "building" part?
> 
> Yeah, I thought the requirement for these scripts was to not include source. build_and_pull_from_git can use custom build and install scripts, so the build scripts can be empty, and the install script can assemble the package folders before they are tar'd up

Looking at our p4 repo, it looks like it was pulled in from commit 6534506e829a489bda78bc5eac5faa34da0a2c51. What we would do it use this tag in the build config and to do a diff against the tag and the one in our repo to produce a patch to apply (which the script will do as well)

lawsonamzn:
Its not an absolute requirement not to include source, especially if its tiny and bespoke.
But its better to not include source if its 'easy' not to...

santorac:
> Looking at our p4 repo, it looks like it was pulled in from commit 6534506e829a489bda78bc5eac5faa34da0a2c51. What we would do it use this tag in the build config and to do a diff against the tag and the one in our repo to produce a patch to apply (which the script will do as well)

Thanks for pointing this out. I'll follow up with @AMZN-koppersr to see if he knows the history here.

lumberyard-employee-dm (approved):

spham-amzn (approved):

AMZN-koppersr:
> > Looking at our p4 repo, it looks like it was pulled in from commit 6534506e829a489bda78bc5eac5faa34da0a2c51. What we would do it use this tag in the build config and to do a diff against the tag and the one in our repo to produce a patch to apply (which the script will do as well)
> 
> Thanks for pointing this out. I'll follow up with @AMZN-koppersr to see if he knows the history here.

There's wasn't a particular motivation behind picking that specific commit as far as I remember. It was part of an effort to replace the modified RapidJSON 1.0.1 version with a clean version of the newer 1.1.0 release. I picked the later commit instead of the exact 1.1.0 tag commit because it had a few more bug fixes. It's ok to stick with that specific commit because it's been proven to work or go to a newer version as there have been a bunch more fixes made to RapidJSON.

lawsonamzn (approved):

package_build_list_host_darwin.json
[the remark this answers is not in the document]
lumberyard-employee-dm: This was from the original revision.
That was fixed in later commits.

package-system/RapidJSON-multiplatform/Readme.txt
```diff
@@ -0,0 +1,3 @@
+RapidJSON 1.1.0
+
+This package was copied from Amazon's 3rdParty source repository. It corresponds to https://github.com/Tencent/rapidjson/releases/tag/v1.1.0
```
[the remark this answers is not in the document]
santorac: This file was removed.

PR #23, opened 2021-08-20
Move 3p AWSNativeSDK linux to github

onecent1101:
## Details
2. Add custom build and install script for linux

Note: Bump the windows and linux package version as the FindAWSNativeSDK need a copyright update

## Testing
Tested locally
Get the same failures as nightly build, the failures are caused by dev branch.
Windows:
https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/O3DE-LY-FORK/detail/LYN-5564/4/pipeline

Linux:
https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/O3DE-LY-FORK/detail/LYN-5564/5/pipeline

Signed-off-by: liug <liug@uf5437f6faf8e57.ant.amazon.com>

### DIFF
--- a/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux
+++ b/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux
@@ -1,0 +1,356 @@
+#
+# All or portions of this file Copyright (c) Amazon.com, Inc. or its affiliates or
+# its licensors.
+#
+# For complete copyright and license terms please see the LICENSE at the root of this
+# distribution (the "License"). All use of this software is governed by the License,
+# or, if provided, by the license below or the license accompanying this file. Do not
+# remove or modify any license notices. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#
+
+include(CMakeParseArguments)
+
+set(AWSNATIVESDK_PACKAGE_NAME AWSNativeSDK)
+
+set(AWS_BASE_PATH ${CMAKE_CURRENT_LIST_DIR}/${AWSNATIVESDK_PACKAGE_NAME})
+
+# Include Path
+set(AWSNATIVESDK_INCLUDE_PATH ${AWS_BASE_PATH}/include)
+
+# Determine the lib path
+if(LY_MONOLITHIC_GAME)
+    set(AWSNATIVE_SDK_LIB_PATH ${AWS_BASE_PATH}/lib/$<IF:$<CONFIG:Debug>,Debug,Release>)
+else()
+    set(AWSNATIVE_SDK_LIB_PATH ${AWS_BASE_PATH}/bin/$<IF:$<CONFIG:Debug>,Debug,Release>)
+endif()
+
+# AWS Compile Definitions
+set(AWSNATIVESDK_COMPILE_DEFINITIONS AWS_CUSTOM_MEMORY_MANAGEMENT PLATFORM_SUPPORTS_AWS_NATIVE_SDK)
+
+set(AWSNATIVESDK_BUILD_DEPENDENCIES
+    curl # The one bundled with the aws sdk in 3rdParty doesn't use the right openssl
+)
+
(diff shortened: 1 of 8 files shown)


package-system/AWSNativeSDK/build_AWSNativeSDK_linux.sh
```diff
@@ -0,0 +1,60 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+echo "Install required library: zlib1g-dev"
+sudo apt install -y zlib1g-dev || (echo "Install required library: zlib1g-dev failed" ; exit 1)
+
+echo "Install required library: libssl-dev"
+sudo apt install -y libssl-dev || (echo "Install required library: libssl-dev failed" ; exit 1)
+
+echo "Install required library: libcurl4-openssl-dev"
+sudo apt install -y libcurl4-openssl-dev || (echo "Install required library: libcurl4-openssl-dev failed" ; exit 1)
```
lawsonamzn, line 15: I dont think we should be sudo-ing in any of these scripts - we intend this to run automatically and unattended at some point, so we can't require admin

a better way (on ubuntu) is to just check if its installed instead, which does NOT require sudo
```
if dpkg-query -W -f'${Status}' "zlib1g-dev" 2>/dev/null | grep -q "ok installed"; then
 echo Required package zlib1g-dev is not installed.
 exit 1
fi;
```

But I think its probably even better not to check and allow the build script to fail.
Maybe print at the top "you need zlib1g-dev, libssl-dev, and libcurl4-openssl-dev in order to compile this"

This is so that if you're not on ubuntu (like you're on redhat) you still have a chance to satisfy its requirements and compile it anyway

onecent1101: sounds good, will do

spham-amzn (approved):

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux
```diff
@@ -0,0 +1,356 @@
...
+# All or portions of this file Copyright (c) Amazon.com, Inc. or its affiliates or
+# its licensors.
+#
+# For complete copyright and license terms please see the LICENSE at the root of this
+# distribution (the "License"). All use of this software is governed by the License,
+# or, if provided, by the license below or the license accompanying this file. Do not
+# remove or modify any license notices. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#
+
```
lawsonamzn, line 11: SHouldn't this be the o3de license?

package-system/AWSNativeSDK/build_AWSNativeSDK_linux.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
lawsonamzn, line 7: Maybe copy this license (minus the bin bash line) to the other file?

lawsonamzn: this also makes me think we probably SHOULD update the package becuase we do want the proper license :(

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux
onecent1101: Will update

package-system/AWSNativeSDK/build_AWSNativeSDK_linux.sh
onecent1101: package update will be done with version update (to 1.9) together after 3p migration.

onecent1101: done

lawsonamzn (approved):

PR #24, opened 2021-08-25
3P Conversion: Squish-ccr

VickyAtAZ:
SPEC-7821 3P Conversion: Squish-ccr

It uses the extras/pull_and_build_from_git.py with template json file to build the library. The main different compare to previous package are:
* Switched to use SSE2 instead of SSE4.1 since the SSE4.1 leads to compression issue with BC6 format (some black pixel blocks) 
* Switched from static library to dynamic library since it helps performance with AP debug build.


package-system/squish-ccr/build_config.json
```diff
@@ -0,0 +1,82 @@
...
+      "Windows":{
+        "Windows": {
+            "custom_cmake_install": true,
+            "cmake_generate_args_release": [
+                "-G",
+                "\"Visual Studio 16 2019\"",
+                "-A",
+                "x64",
+                "-T",
+                "host=x64",
```
hershey5045, line 27: These aren't required anymore for VS2019 (architecture and toolset args)

VickyAtAZ:
Using a new PR https://github.com/o3de/3p-package-source/pull/26 to replace this one

PR #32, opened 2021-08-27
Added pdbs for python windows in order to allow python/c++ debugging on VS 

AMZN-AlexOteiza:


AMZN-AlexOteiza:
Related PR: https://github.com/o3de/o3de/pull/3661

PR #34, opened 2021-08-30
Updates the zlib package to only build statically

lawsonamzn:
Update zlib to build statically only
Also adds a FindZLIB (capital letters, in a subfolder) that contains no o3de macros and thus can be used by any 3p that wants zlib, making it MUCH simpler to use our zlib in other 3p's that depend on it. 


package-system/zlib/build_zlib_linux.sh
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+cmake -S temp/src -B temp/build -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DCMAKE_C_FLAGS=-fPIC -DSKIP_INSTALL_FILES=YES
```
spham-amzn, line 9: Should add a status code check after the generation as well

```
if [ $? -ne 0 ]
then
    echo "Error generating build"
    exit $?
fi
```

package-system/zlib/install_zlib_linux.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+
+OUT_PATH=$TARGET_INSTALL_ROOT
+SRC_PATH=temp/src
+BLD_PATH=temp/build
+
+mkdir -p $OUT_PATH
+
+cp -f $SRC_PATH/LICENSE $OUT_PATH/
+mkdir -p $OUT_PATH/lib
```
spham-amzn, line 16: same error check should go to the .sh scripts as well.

package-system/zlib/build_zlib_linux.sh
lawsonamzn: ok. I can add this without actually having to rebuild the package, thankfully.

lawsonamzn: Done, although, I've changed it to exit 1 instead of $? becuase by the time it would call exit, the ? is no longer valid.

spham-amzn: ah thats right, echo would reset it

spham-amzn (approved):

package-system/zlib/install_zlib_linux.sh
lawsonamzn: Added these too.

lawsonamzn: note that I tested by intentionally making a misspelling so that it fails.  (Temporariily!)

spham-amzn (approved):

amzn-changml (approved):

PR #35, opened 2021-08-31
Update script to support git override

onecent1101:
## Testing
Tested locally with platform override, it works well

Signed-off-by: liug <liug@uf5437f6faf8e57.ant.amazon.com>

### DIFF
--- a/Scripts/extras/pull_and_build_from_git.py
+++ b/Scripts/extras/pull_and_build_from_git.py
@@ -26,8 +26,6 @@ The build configuration (build_config.json) accepts keys that are root level onl
 either global or target platform specific. Root level only keys are keys that define the project and cannot
 be different by platform, and all are required. The keys are:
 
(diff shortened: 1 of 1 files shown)


lawsonamzn:
This would allow overriding the git tag to use per platform, right?

spham-amzn:
> 
> 
> This would allow overriding the git tag to use per platform, right? If so consider updating the docs near the top of the file. it has a section for "stuff that can only be in the root" vs "stuff that can be deeper in the doc"

Yes, that is correct

onecent1101:
> This would allow overriding the git tag to use per platform, right? If so consider updating the docs near the top of the file. it has a section for "stuff that can only be in the root" vs "stuff that can be deeper in the doc"

Done

lawsonamzn (approved):

junbo75 (approved):

PR #36, opened 2021-09-01
[main] adding RapidXML package

AMZN-ScottR:
Signed-off-by: AMZN-ScottR <24445312+AMZN-ScottR@users.noreply.github.com>


AMZN-alexpete (approved):

AMZN-ScottR:
> How hard would it be to just download the source from the github repo?
> 
> It seems like we could perform actions that we do with the [rapidjson](https://github.com/o3de/3p-package-source/commit/2042a465144093c9d190a30b9a93d71338c3a3dc) 3p-package-source to download it from git

There unfortunately isn't a main GH repo for RapidXML, its primary distribution is a zip from SourceForge.

lawsonamzn (approved):
Yes.  At some point later we could upgrade pull_and_build_From_git or something to allow zip files or create a zip downloader but for such a tiny lib its likely not worth it.

PR #40, opened 2021-09-09
Added libpng

santorac:
Added libpng so we can make o3de use this instead of OpenImageIO, as libpng is a much leaner dependency and is all that's needed for the runtime.

Note that platforms other than Windows have not been tested yet; will get that done before committing.


package-system/libpng/Findlibpng.cmake
```diff
@@ -0,0 +1,42 @@
...
+elseif(${PAL_PLATFORM_NAME} STREQUAL "Windows")
+    set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/libpng16_staticd.lib)
+    set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/libpng16_static.lib)
+elseif(${PAL_PLATFORM_NAME} STREQUAL "Mac")
+    set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/libpng16.a)
+    set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/libpng16.a)
+elseif(${PAL_PLATFORM_NAME} STREQUAL "iOS")
+    set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/libpng16.a)
+    set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/libpng16.a)
+endif()
```
rgba16f, line 31: Could this be changed to just 
```
if(${PAL_PLATFORM_NAME} STREQUAL "Windows")
    set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/libpng16_staticd.lib)
    set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/libpng16_static.lib)
else()
    set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/libpng16.a)
    set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/libpng16.a)
endif()
```

package-system/libpng/build_config.json
```diff
@@ -0,0 +1,49 @@
...
+    "package_url": "http://www.libpng.org/",
+    "package_license": "libpng-2.0",
+    "package_license_file": "LICENSE",
+    "cmake_find_source": "Findlibpng.cmake",
+    "cmake_find_target": "Findlibpng.cmake",
+    "Platforms": {
+        "Windows": {
+            "Windows": {
+                "cmake_generate_args": [
+                    "-DZLIB_ROOT=../../zlib-windows/zlib",
```
rgba16f, line 15: Does this require you to have manually build zlib first?

lumberyard-employee-dm: @nemerle  did bring up the possibility of using the [std_image_write.h](https://github.com/nothings/stb/blob/master/stb_image_write.h) header for writing pngs in this [comment](https://github.com/o3de/o3de/pull/4006#issuecomment-915940117)

That would avoid the explicit dependency on zlib.

lawsonamzn: depending on zlib like this realyl isn't a big deal though.  We build it and the new one as of last week is cmake findzlib compatible.   The only issue is having to build it first.

nemerle: My suggestion was mostly related to resulting object-file sizes :smile:

rgba16f: The only side-effect is that if you forget to build zlib locally some platforms will use the system library zlib copy. Mac & iOS both work without building zlib.

lumberyard-employee-dm (approved):

package-system/libpng/build_config.json
lawsonamzn: I see.  We've actually had issues with that before.   For example, on mac, XCode ships with a LZMA library, so libtiff would detect and use it automatically. But, because it doesn't actually specify to _link_ to it, you end up with a linker error about missing lzma symbols.  I personally prefer to control the dependencies exactly (either use O3DE's real library, or don't use any library version at all, with the latter being the better choice if possible.)

santorac: I stated in the readme.txt that you have to build the o3de zlib package locally first. This is how I built the Windows package. Is there a way we can prevent libpng from using some other zlib it finds? (I can investigate this if you all don't know off hand).

lawsonamzn: it kind of depends.  I assume libpng uses cmake to build and not a custom make, if so, cmake accepts this

(libname)_DIR=(folder to look) to override where it looks for 3p

so, eg, adding -DZLIB_DIR=../../zlib-windows/zlib would do it.  Note that we want the 'zlib' subfolder, not the root, so that it uses the FindZLIB.cmake file in there, not the o3de "Findzlib.cmake" in the root.

package-system/libpng/build_config.json
```diff
@@ -0,0 +1,67 @@
...
+                    "-DPNG_DEBUG=OFF"
+                ]
+            },
+            "iOS": {
+                "cmake_generate_args": [
+                    "-DZLIB_ROOT=../../zlib-ios/zlib",
+                    "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW",
+                    "-DPNG_SHARED=OFF",
+                    "-DPNG_TESTS=OFF",
+                    "-DPNG_DEBUG=OFF",
```
lawsonamzn, line 48: I'm surprised this works withotu also specifying -DCMAKE_MACOSX_BUNDLE=OFF 
But if it does, great!

lawsonamzn (approved):

package-system/libpng/Findlibpng.cmake
santorac: Maybe, but I was following the pattern shown in other "Find" files in the package system.

package-system/libpng/build_config.json
santorac: Okay, since this is using CMP0074=NEW and -DZLIB_ROOT=../../zlib-windows/zlib it sounds like we have what we need.

package-system/libpng/Findlibpng.cmake
```diff
@@ -0,0 +1,42 @@
...
+
+# we set it to a generator expression for multi-config situations:
+set(libpng_LIBRARY                   "$<$<CONFIG:profile>:${libpng_LIBRARY_RELEASE}>")
+set(libpng_LIBRARY ${libpng_LIBRARY} "$<$<CONFIG:Release>:${libpng_LIBRARY_RELEASE}>")
+set(libpng_LIBRARY ${libpng_LIBRARY} "$<$<CONFIG:Debug>:${libpng_LIBRARY_DEBUG}>")
+
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+ly_target_include_system_directories(TARGET ${TARGET_WITH_NAMESPACE} INTERFACE ${${MY_NAME}_INCLUDE_DIR})
+target_link_libraries(${TARGET_WITH_NAMESPACE} INTERFACE ${libpng_LIBRARY})
```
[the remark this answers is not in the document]
lumberyard-employee-dm:
```suggestion
add_library(${TARGET_WITH_NAMESPACE} STATIC IMPORTED GLOBAL)
ly_target_include_system_directories(TARGET ${TARGET_WITH_NAMESPACE} INTERFACE ${${MY_NAME}_INCLUDE_DIR})
set_target_properties(${TARGET_WITH_NAMESPACE}
    PROPERTIES
        IMPORTED_LOCATION_DEBUG ${${MY_NAME}_LIBRARY_DEBUG}
        IMPORTED_LOCATION_PROFILE ${${MY_NAME}_LIBRARY_RELEASE}
        IMPORTED_LOCATION_RELEASE ${${MY_NAME}_LIBRARY_RELEASE}
)
target_link_libraries(${TARGET_WITH_NAMESPACE} INTERFACE 3rdParty::zlib)
```

package-system/libpng/Findlibpng.cmake
```diff
@@ -35,8 +35,15 @@ set(libpng_LIBRARY                   "$<$<CONFIG:profile>:${libpng_LIBRARY_RELEA
...
-add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+add_library(${TARGET_WITH_NAMESPACE} STATIC IMPORTED GLOBAL)
 ly_target_include_system_directories(TARGET ${TARGET_WITH_NAMESPACE} INTERFACE ${${MY_NAME}_INCLUDE_DIR})
-target_link_libraries(${TARGET_WITH_NAMESPACE} INTERFACE ${libpng_LIBRARY})
+# "Because libpng depends on the zlib, zlib needs to be added as a link dependency of the libpng target, which must be a STATIC IMPORTED target with an IMPORTED_LOCATION_$<CONFIG> set to be library files."
+set_target_properties(${TARGET_WITH_NAMESPACE}
+    PROPERTIES
+        IMPORTED_LOCATION_DEBUG ${${MY_NAME}_LIBRARY_DEBUG}
+        IMPORTED_LOCATION_PROFILE ${${MY_NAME}_LIBRARY_RELEASE}
+        IMPORTED_LOCATION_RELEASE ${${MY_NAME}_LIBRARY_RELEASE}
```
lumberyard-employee-dm, line 45: After discussion we could collapse the `IMPORTED_LOCATION_PROFILE` and `IMPORTED_LOCATION_RELEASE` to a single `IMPORTED_LOCATION` property that references the Release library

```suggestion
        IMPORTED_LOCATION ${${MY_NAME}_LIBRARY_RELEASE}
```

package-system/libpng/Findlibpng.cmake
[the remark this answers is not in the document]
lawsonamzn: Why specify _DEBUG, _PROFILE, and _RELEASE whe you could just specify 
IMPORTED_LOCATION (default to use)
IMPORTED_LOCATION_xxxxx <-- if you wanna override the default, which not all packages need to do

PR #41, opened 2021-09-09
add freetype library build scripts

rgba16f:


package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+import sys
+
+parser = argparse.ArgumentParser(description='Builds this package')
+parser.add_argument('--platform', default=platform.system().lower(), required=False, help=f'Platform to build (defaults to \"{platform.system().lower()}\")')
+args = parser.parse_args()
+
+this_dir = os.path.dirname(os.path.realpath(__file__))
+cmake_scripts_path = os.path.abspath(os.path.join(this_dir, '../../Scripts/cmake'))
+
+ly_3rdparty_path = os.getenv('LY_3RDPARTY_PATH')
```
spham-amzn, line 22: is this variable used?

rgba16f: It appears not. This file was originally in p4, and that line is from the previous version. If I need to do another package revision I will remove it.

package-system/freetype/freetype.patch
```diff
@@ -0,0 +1,28 @@
...
+@@ -405,7 +405,7 @@ if (WIN32)
+ endif ()
+ 
+ if (BUILD_SHARED_LIBS)
+-  set_target_properties(freetype PROPERTIES
++set_target_properties(freetype PROPERTIES
+     VERSION ${LIBRARY_VERSION}
+     SOVERSION ${LIBRARY_SOVERSION})
+ endif ()
+diff --git a/include/ft2build.h b/include/ft2build.h
```
lumberyard-employee-dm, line 14: All this part of the patch does is remove a space.
We can the patch portion for the CMakeLists.txt from the patch

package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+        '-G', 'Ninja Multi-Config',
+        f'-DCMAKE_TOOLCHAIN_FILE={cmake_scripts_path}/Platform/Android/Toolchain_android.cmake',
+        '-DANDROID_ABI=arm64-v8a',
+        '-DANDROID_ARM_MODE=arm',
+        '-DANDROID_ARM_NEON=FALSE',
+        '-DANDROID_NATIVE_API_LEVEL=21',
+        f'-DLY_NDK_DIR={ly_android_sdk_path}/ndk/22.0.7026061',
+        '-DPACKAGE_PLATFORM=android',
+        '-DCMAKE_POSITION_INDEPENDENT_CODE=ON'
+    ], [], 'debug', 'release') # Android needs to have ninja in the path
```
lumberyard-employee-dm, line 59: Why lowercase `debug` and `release` on Android?

package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+        '-G', 'Ninja Multi-Config',
+        '-DCMAKE_C_COMPILER=clang-6.0', 
+        '-DCMAKE_CXX_COMPILER=clang++-6.0',
+        '-DCMAKE_POSITION_INDEPENDENT_CODE=ON'
+    ], [], 'Debug', 'Release'),
+    'windows'   : ([
+        '-G', 'Visual Studio 16 2019',
+        '-Ax64', '-Thost=x64'
+    ], [], 'Debug', 'Release'),
+    'android'   : ([
```
lumberyard-employee-dm, line 49: We can get away with only Release configurations on mac/ios/ linux and android since we have no need to debug the freetype library itself.

package-system/freetype/build_package_image.py
rgba16f: That appears to be how gradle names the builds. These were determined by what it took for the build to work on each platform.

package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+    sys.exit(result_value.returncode)
+
+# build debug
+build_call =['cmake', '--build', build_dir, '--config', debug_build_name, '--target', 'install']
+if cmake_build:
+    build_call += cmake_build
+print(build_call)
+result_value = subprocess.run(build_call, shell=False, cwd=script_dir)
+if result_value.returncode != 0:
+    sys.exit(result_value.returncode)
```
lumberyard-employee-dm, line 95: I think we should add a check to see if the debug_build_name is not "truthy" before attempting to build debug.

```suggestion
# build debug
if debug_build_name:
    build_call =['cmake', '--build', build_dir, '--config', debug_build_name, '--target', 'install']
    if cmake_build:
        build_call += cmake_build
    print(build_call)
    result_value = subprocess.run(build_call, shell=False, cwd=script_dir)
    if result_value.returncode != 0:
        sys.exit(result_value.returncode)
```

That would allow conditional compliation of the debug configuration if it exists.

package-system/freetype/build_package_image.py
rgba16f: That is correct but I'd rather the script was uniform. Changing it would require retesting every platform and build again.

lumberyard-employee-dm: Performance wise, debug builds aren't always efficient for these types of libraries.
This was noticed in the past with debug version of squish-ccr causing processing of specific tiff assets to timeout in debug after an hour on Windows, while the release version version was able to finish asset processing in a timely manner.

lumberyard-employee-dm (approved):

package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+
+parser = argparse.ArgumentParser(description='Builds this package')
+parser.add_argument('--platform', default=platform.system().lower(), required=False, help=f'Platform to build (defaults to \"{platform.system().lower()}\")')
+args = parser.parse_args()
+
+this_dir = os.path.dirname(os.path.realpath(__file__))
+cmake_scripts_path = os.path.abspath(os.path.join(this_dir, '../../Scripts/cmake'))
+
+ly_3rdparty_path = os.getenv('LY_3RDPARTY_PATH')
+ly_android_sdk_path = os.getenv('LY_ANDROID_SDK_PATH');
```
spham-amzn, line 23: nit: Should let the user know that the Android SDK path was not set here, or the path is invalid, so that they can correct it.

lawsonamzn (approved):

package-system/freetype/build_package_image.py
```diff
@@ -0,0 +1,102 @@
...
+                 '-DCMAKE_DISABLE_FIND_PACKAGE_BZip2=TRUE',
+                 '-DCMAKE_DISABLE_FIND_PACKAGE_PNG=TRUE',
+                 '-DCMAKE_DISABLE_FIND_PACKAGE_HarfBuzz=TRUE',
+                 '-DCMAKE_DISABLE_FIND_PACKAGE_BrotliDec=TRUE']
+if cmake_generation:
+    generate_call += cmake_generation
+print(f"Cmake command '{generate_call}'")
+result_value = subprocess.run(generate_call, shell=False, cwd=script_dir)
+if result_value.returncode != 0:
+    sys.exit(result_value.returncode)
```
burelc-amzn, line 86: Nit: You can use `subprocess.check_call` here instead. And `shell=False` is already the default value.

burelc-amzn (approved):

PR #42, opened 2021-09-10
Move 3p AWSNativeSDK ios to github

onecent1101:
## Details
2. Add custom build and install script for ios

Note: Bump the ios package version as the FindAWSNativeSDK need a copyright update

## Testing
Tested locally
AR testing https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/O3DE-LY-FORK/detail/LYN-6779/4/pipeline/

Signed-off-by: liug <liug@amazon.com>

### DIFF
--- a/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.iOS
+++ b/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.iOS
@@ -1,0 +1,312 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+
+include(CMakeParseArguments)
+
+set(AWSNATIVESDK_PACKAGE_NAME AWSNativeSDK)
+
+set(AWS_BASE_PATH ${CMAKE_CURRENT_LIST_DIR}/${AWSNATIVESDK_PACKAGE_NAME})
+
+# Include Path
+set(AWSNATIVESDK_INCLUDE_PATH ${AWS_BASE_PATH}/include)
+
+# Only static libs are supported for this revision
+set(AWSNATIVE_SDK_LIB_PATH ${AWS_BASE_PATH}/lib/$<IF:$<CONFIG:Debug>,Debug,Release>)
+
+# AWS Compile Definitions
+set(AWSNATIVESDK_COMPILE_DEFINITIONS AWS_CUSTOM_MEMORY_MANAGEMENT PLATFORM_SUPPORTS_AWS_NATIVE_SDK)
+
+find_library(SECURITY_FRAMEWORK Security)
+set(AWSNATIVESDK_BUILD_DEPENDENCIES ${SECURITY_FRAMEWORK})
+
+# Helper function to define individual AWSNativeSDK Libraries
+function(ly_declare_aws_library)
+
+    set(options)
+    set(oneValueArgs NAME LIB_FILE)
+    set(multiValueArgs BUILD_DEPENDENCIES)
+    
+    cmake_parse_arguments(ly_declare_aws_library "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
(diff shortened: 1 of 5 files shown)


lawsonamzn (approved):
Looks good to me.  Isn't this already there?

onecent1101:
> 
> 
> Looks good to me. Isn't this already there?

Next step is to upload rev4 version to dev s3, verify through AR build, promote to prod s3 and send out engine 3rdparty code version update

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.iOS
```diff
@@ -0,0 +1,312 @@
...
+include(CMakeParseArguments)
+
+set(AWSNATIVESDK_PACKAGE_NAME AWSNativeSDK)
+
+set(AWS_BASE_PATH ${CMAKE_CURRENT_LIST_DIR}/${AWSNATIVESDK_PACKAGE_NAME})
+
+# Include Path
+set(AWSNATIVESDK_INCLUDE_PATH ${AWS_BASE_PATH}/include)
+
+# Only static libs are supported for this revision
```
junbo75, line 17: Curious: Is there a plan to add dynamic libraries later or is it not supported on iOS at all?

package-system/AWSNativeSDK/build_AWSNativeSDK_ios.sh
```diff
@@ -0,0 +1,95 @@
...
+
+    echo "CMake Build $build_type Static to $bld_path/${build_type}_Static"
+    cmake --build "$bld_path/${build_type}_Static" --config $build_type -j 12 || (echo "CMake Build $build_type Static to $bld_path/${build_type}_Static failed" ; exit 1)
+}
+
+make_configure_and_build_curl() {
+  rm -rf "temp/curl"*
+
+  echo "Downloading Curl 7.65.3"
+  (cd temp && curl -o curl-7.65.3.zip "https://curl.se/download/curl-7.65.3.zip") || exit 1
```
junbo75, line 41: Why not clone from Github like what your implementation in the cmake_configure_and_build_curl() function below?

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.iOS
onecent1101: no unless there is strong requirement, as that is how we setup previously (not sure whether there is any limitation here)

package-system/AWSNativeSDK/build_AWSNativeSDK_ios.sh
onecent1101: source cloned from github repo doesn't support make build directly.

junbo75 (approved):

package-system/AWSNativeSDK/build_AWSNativeSDK_ios.sh
```diff
@@ -0,0 +1,95 @@
...
+
+  (cd temp/curl-7.65.3 && ./configure --disable-shared --enable-static --enable-ipv6 --with-secure-transport --host="arm-apple-darwin" --prefix=$(pwd)/../curl_install) || exit 1
+  (cd temp/curl-7.65.3 && make) || exit 1
+  (cd temp/curl-7.65.3 && make install) || exit 1
+
+  export CFLAGS=$EXISTING_CFLAGS
+  EXISTING_CFLAGS=
+}
+
+# TODO: curl cmake is poorly maintained by community, switch to cmake when it is ready
```
lmbr-pip, line 57: This is an unsubstantiated comment, theres no justification for this statement given. Also we are using CMake so unsure what task actually is. How does anyone reading this take action? What does 'poorly maintained' mean in this context, what is the cmake action?

lmbr-pip (commented):
LG

IMHO 1 comment needs an update.

package-system/AWSNativeSDK/build_AWSNativeSDK_ios.sh
onecent1101: actually we are not using cmake for curl build, instead we are using makefile.
the comment "cmake is poorly maintained by community" is copied from cmake build result, as cmake failed when I tried to use it to build.
I can remove the whole function, left it to have a reference we can use later.

lawsonamzn (approved):

PR #43, opened 2021-09-16
Move 3p AWSNativeSDK android and upgrade major version to github

onecent1101:
## Details
Add custom build and install script for android, and bump android package major version
Here we do package major version bump for  android first as new aws sdk cpp helps to solve the sdk dependency build issue, other platforms major version bump will be tracked separately

## Testing
Tested locally

AR testing https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/O3DE-LY-FORK/detail/LYN-6779/2/pipeline

Signed-off-by: Liu <liug@amazon.com>

### DIFF
--- a/package-system/AWSNativeSDK/build_AWSNativeSDK_android.cmd
+++ b/package-system/AWSNativeSDK/build_AWSNativeSDK_android.cmd
@@ -1,0 +1,81 @@
+@echo off
+REM 
+REM Copyright (c) Contributors to the Open 3D Engine Project.
+REM For complete copyright and license terms please see the LICENSE at the root of this distribution.
+REM 
+REM SPDX-License-Identifier: Apache-2.0 OR MIT
+REM 
+
+SET SRC_PATH=temp\src
+SET BLD_PATH=temp\build
+
+IF "%ANDROID_NDK_ROOT%"=="" (
+    ECHO "Required envrironment variable ANDROID_NDK_ROOT is missing, please set it to local android ndk directory"
+    exit /b 1
+)
+
+REM Debug Shared
+call:ConfigureAndBuild Debug Shared
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
+)
+
+REM Debug Static
+call:ConfigureAndBuild Debug Static
+IF %ERRORLEVEL% NEQ 0 (
+    exit /b 1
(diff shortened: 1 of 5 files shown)


junbo75 (approved):
LG. The only concern is that Android SDK has a different major version, which makes it inconsistent with the other platforms. Feels like we should finish the update for other platforms as soon as possible once this change is checked in.

onecent1101:
> LG. The only concern is that Android SDK has a different major version, which makes it inconsistent with the other platforms. Feels like we should finish the update for other platforms as soon as possible once this change is checked in.

yes, but thanks to aws sdk, most of them are backward compatible

package-system/AWSNativeSDK/build_AWSNativeSDK_android.cmd
```diff
@@ -0,0 +1,81 @@
...
+    SET BUILD_SHARED=ON
+)
+ECHO "CMake Configure %BUILD_TYPE% %LIB_TYPE%"
+call cmake -S %SRC_PATH% -B %BLD_PATH%\%BUILD_TYPE%_%LIB_TYPE% ^
+           -G Ninja ^
+           -DNDK_DIR="%ANDROID_NDK_ROOT%" ^
+           -DBUILD_SHARED_LIBS=%BUILD_SHARED% ^
+           -DCMAKE_BUILD_TYPE="%BUILD_TYPE%" ^
+           -DTARGET_ARCH=ANDROID ^
+           -DANDROID_NATIVE_API_LEVEL=21 ^
```
lmbr-pip, line 58: I wonder if theres some overall value to track here, rather than each 3P defining its only minimal Android version?

lmbr-pip (approved):
LG.

Nit: This should have a title like "Migrate and update 3P AWSNativeSDK for Android to Github" and major new version should be in PR description.

package-system/AWSNativeSDK/build_AWSNativeSDK_android.cmd
onecent1101: not sure, but it is minimal level required by aws sdk cpp https://github.com/aws/aws-sdk-cpp/blob/main/Docs/CMake_Parameters.md#android_native_api_level

PR #44, opened 2021-09-16
SPEC-7799  3P Conversion: ASTCEncoder

VickyAtAZ:
Convert to using build script for 3rdparty astc encoder library.
Also, instead of only contains executable, the library also includes static libraries for using its APIs.
The executable is included in the package but it's not added for runtime since we don't need it for runtime anymore. 
The changes in patch file are required to address linking issues for different platform. 

The o3de change to use this new library is in  this [PR](https://github.com/o3de/o3de/pull/4157)

The library was tested on windows/Mac/Linux with both profile and debug builds.


package-system/astc-encoder/Findastc-encoder.cmake.template
```diff
@@ -0,0 +1,34 @@
...
+endif()
+
+# declare the target so that others can 'depend on it'
+add_library($${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+
+# add include directory
+ly_target_include_system_directories(TARGET $${TARGET_WITH_NAMESPACE} INTERFACE $${$${LIB_NAME}_INCLUDE_DIR})
+
+# add static library for linking
+target_link_libraries($${TARGET_WITH_NAMESPACE} INTERFACE $${$${LIB_NAME}_LIBRARY})
```
lumberyard-employee-dm, line 32: So there is a new way to define the STATIC library dependencies for a 3rdParty library, that involes making a `STATIC IMPORTED GLOBAL` target which can be used to associate the static library with the `3rdParty::astc-encoder` target, instead of having the `3rdParty::astc-encoder` target depend on path to the static library.

This involves setting the IMPORTED_LOCATION property for the library target
```suggestion
if ($${PAL_PLATFORM_NAME} STREQUAL "Windows")
    set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBRARY_DIR}/Releaseastcenc-sse2-static.lib)
    set($${LIB_NAME}_LIBRARY_DEBUG $${$${LIB_NAME}_LIBRARY_DIR}/Debug/astcenc-sse2-static.lib
elseif ($${PAL_PLATFORM_NAME} STREQUAL "Linux")
    set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBRARY_DIR}/libastcenc-sse2-static.a)
    set($${LIB_NAME}_LIBRARY_DEBUG $${$${LIB_NAME}_LIBRARY_RELEASE})
else()
    set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBRARY_DIR}/libastcenc-native-static.a)
    set($${LIB_NAME}_LIBRARY_DEBUG $${$${LIB_NAME}_LIBRARY_RELEASE})
endif()

# declare the target so that others can 'depend on it'
add_library($${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)

# add include directory
ly_target_include_system_directories(TARGET $${TARGET_WITH_NAMESPACE} INTERFACE $${$${LIB_NAME}_INCLUDE_DIR})

# add static library as "output" IMPORTED_LOCATION for the target
set_target_properties($${TARGET_WITH_NAMESPACE}
    PROPERTIES
        IMPORTED_LOCATION_DEBUG $${$${LIB_NAME}_LIBRARY_DEBUG}
        IMPORTED_LOCATION $${$${LIB_NAME}_LIBRARY_RELEASE}
)
```

package-system/astc-encoder/install_linux.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# copy LICENSE.txt and header file
+cp -f temp/src/LICENSE.txt $TARGET_INSTALL_ROOT/
+cp -f temp/src/Source/astcenc.h $INCLUDE_PATH/
+
+BUILD_PATH=temp/build/Source
+
+# copy static lib and executable
+cp -f $BUILD_PATH/astcenc-sse2 $BIN_PATH/
+cp -f $BUILD_PATH/libastcenc-sse2-static.a $BIN_PATH/
```
lawsonamzn, line 22: It may be useful here to fail the script if the cp fails.
In unix bash its prety easy
```
(command) || exit $?
```

like
```
cp -f temp/src/LICENSE.txt $TARGET_INSTALL_ROOT/ || exit $?
```
will cause it to either succeed the copy OR exit with the exit code of cp.

Note that this file does not go into the package, so changing it would not require a re-upload.

package-system/astc-encoder/install_windows.cmd
```diff
@@ -0,0 +1,33 @@
...
+
+SET BIN_PATH=%TARGET_INSTALL_ROOT%\bin
+SET INCLUDE_PATH=%TARGET_INSTALL_ROOT%\include
+
+mkdir %INCLUDE_PATH%
+mkdir %BIN_PATH%
+mkdir %BIN_PATH%\Release
+mkdir %BIN_PATH%\Debug
+
+copy /Y temp\src\LICENSE.TXT %TARGET_INSTALL_ROOT%\
```
lawsonamzn, line 22: Its probably also worth doing the same thing here 
```
@if %errorlevel% NEQ 0 ( exit /b 1 )
```
except its windows

package-system/astc-encoder/Findastc-encoder.cmake.template
VickyAtAZ: It's good to know. 
I would prefer to keep it as it is since this change would effect all the packages which requires too many follow up validation which I don't think worth it. 
How do you think?

lumberyard-employee-dm: Well if we don't ever do this in any PRs of converting a 3rdParty library now, then no one would ever do it.

Quick question for @VickyAtAZ:
What was your process for adding the ASTCEncoder `Findastc-encoder.template.cmake` script?
Did you take another 3p package .cmake script and adapt that script to work with the ASTCEncoder?
Or did you use the instructions in the 3rdParty runbook?

I am just trying to get a feel for the process that developers are using to convert 3rdParty packages.

VickyAtAZ: I copied one from another script (squish-ccr in this case which I created few weeks ago ). 
I was looking at the run book when I was creating squish-ccr script. And the change about set_target_properties was not there yet.

lawsonamzn (approved):
Looks ok to me.  We may have to go back over all our  libraries and clean up CMake Find Files in the future, but I can't justify holding this back.

package-system/astc-encoder/Findastc-encoder.cmake.template
VickyAtAZ: > 
> 
> Well if we don't ever do this in any PRs of converting a 3rdParty library now, then no one would ever do it.
> 
Unless the majority of the existing scripts are converted, this won't be meaningful. 
Are we going to ask engineers to convert each script individually? I don't think it's the best of using engineers' time. 
If this API is really matters, then we should have a batch converting instead one by one. 

Anyway, I don't think this is necessary for this one script especially it doesn't have any risk of compilation order issue.

VickyAtAZ: I will update the script.

package-system/astc-encoder/Findastc-encoder.cmake.template
```diff
@@ -0,0 +1,40 @@
...
+elseif ($${PAL_PLATFORM_NAME} STREQUAL "Linux")
+    set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBRARY_DIR}/libastcenc-sse2-static.a)
+    set($${LIB_NAME}_LIBRARY_DEBUG $${$${LIB_NAME}_LIBRARY_RELEASE})
+else()
+    set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBRARY_DIR}/libastcenc-native-static.a)
+    set($${LIB_NAME}_LIBRARY_DEBUG $${$${LIB_NAME}_LIBRARY_RELEASE})
+endif()
+
+# declare the target so that others can 'depend on it'
+add_library($${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
```
lumberyard-employee-dm, line 28: Actually the library type needs to be changed to `STATIC` in this case.

lumberyard-employee-dm (approved):

PR #45, opened 2021-09-22
MCPP: replaced 'git apply' with 'patch.py'

galibzon:
MCPP was failing to compile because it was using 'git apply'
to apply the required patch that contains the O3DE code changes.

'git apply' fails silently when it runs under a repository that is not
related with the patch.

To fix the patching issue found 'patch.py' and added a copy of it
into 'Scripts/extras/' folder.

patch.py is installable via 'pip install patch' but it's no supposed
to be used as an API, instead it's supposed to be used as an
application.

'mcpp/get_and_build_mcpp.py' invokes 'patch.py' as a python application
to apply the patch. This method is compatible for Windows, Linux and
Mac.

Technically speaking, Linux and Mac could use the 'patch -p1' command,
but that command is not available on Windows. This is why I added
'patch.py' to the code base.

Additionally added checking for the x64 version of the Visual Studio
Command Prompt which by default is x86. This will force the user
to manually run the x64 version. The x86 version will compile but it
will be incompatible with O3DE.

### DIFF
--- a/Scripts/extras/patch.py
+++ b/Scripts/extras/patch.py
@@ -1,0 +1,1203 @@
+#!/usr/bin/env python
+"""
+    Patch utility to apply unified diffs
+
+    Brute-force line-by-line non-recursive parsing 
+
+    Copyright (c) 2008-2016 anatoly techtonik
+    Available under the terms of MIT license
+
+"""
+from __future__ import print_function
+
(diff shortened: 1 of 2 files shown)


galibzon:
Closing, as bringing another python module would incur in getting legal approval, etc. Fixable with  creating a local git repository before calling git apply.

PR #46, opened 2021-09-23
MCPP: fixed 'git apply' issues on Windows

galibzon:
MCPP was failing to compile because it was using 'git apply'
to apply the required patch that contains the O3DE code changes.

'git apply' fails silently when it runs outside of a git repository.
To avoid this problem 'get_and_build_mcpp.py' now creates a local
git repository before calling 'git apply patch'

It was found that 'git apply' would fail in Windows because of the
LF to CRLF auto conversion. To circumvent this issue , added
.gitattributes file to force the patch file to be fetched without CRLF,
keeping its original LF characters. Also the local git repo is created
with ['git', 'config', '--local', 'core.autocrlf', 'false'] and
the patching works successfully.

Additionally added checking for the x64 version of the Visual Studio
Command Prompt which by default is x86. This will force the user
to manually run the x64 version. The x86 version will compile but it
will be incompatible with O3DE.

### DIFF
--- a/package-system/mcpp/get_and_build_mcpp.py
+++ b/package-system/mcpp/get_and_build_mcpp.py
@@ -52,6 +52,16 @@ elif platform.system() == 'Windows':
             print("This script must be run from a visual studio command prompt, or the visual studio command line"
                   " environments must be set")
             exit(1)
+    # Check it's running under x64 build environment.
+    vs_target_arch = os.environ.get('VSCMD_ARG_TGT_ARCH')
+    if vs_target_arch is None:
+        print("Couldn't read the environment variable 'VSCMD_ARG_TGT_ARCH'. This script must be run from a x64 visual studio command prompt, or the visual studio command line"
+              " environments must be set")
+        exit(1)
+    if vs_target_arch != 'x64':
+        print("This script must be run from a x64 visual studio command prompt, or the visual studio command line"
+              " environments must be set")
+        exit(1)
 else:
     assert False, "Invalid platform"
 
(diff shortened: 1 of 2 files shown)


package-system/mcpp/get_and_build_mcpp.py
```diff
@@ -122,6 +132,43 @@ def extract_tarfile(temp_folder):
...
+    to not autoconvert LF to CRLF because it causes the patching
+    to fail as 'git apply' is very picky about that.
+    2- You may notice that there's a .gitattributes file that makes sure
+    the patch file remains with LF when fetched from the repo.
+    3- It was also found that 'git apply' also failed if the patch had CRLF
+    AND the local git repo also had CRLF.
+    """
+    pristine_source_path = str((temp_folder / SOURCE_NAME).resolve())
+    git_cmds = [
+        ['git', 'init'],
```
spham-amzn, line 150: FYI: Since you are running the execute_cmd with shell=True, you do not have to separate the arguments out into an array. The execute_cmd will just join them into a singular statement because thats how executing a process through a shell works. 

So git_cmds can just look like:

```
git_cmds = [
   'git init',
   'git config --local core.eol lf',
   'git config --local core.autocrlf false',
...
etc
```

spham-amzn (approved):

package-system/mcpp/get_and_build_mcpp.py
galibzon: Thanks for the tip, although it's the same code from "apply_patch()".

lawsonamzn (approved):
Looking good for now.  Although in the future we might want to revisit this and use vswhere to find/activate the vs env, similar to how we build python.

PR #47, opened 2021-09-23
[LYN-4620] Upgrade Mac AWSNativeSDK to 1.9.50

onecent1101:
## Details
Mac AWSNativeSDK major version update

## Testing
Tested locally

Signed-off-by: liug <liug@amazon.com>

### DIFF
--- a/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
+++ b/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
@@ -90,89 +90,68 @@ function(ly_declare_aws_library)
 endfunction()
 
 
-#### CRT ####
+#### Core ####
+set(CORE_STATIC_LIBS
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-crt-cpp.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-auth.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-mqtt.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-http.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-event-stream.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-io.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-checksums.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-compression.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-cal.a
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-common.a
+)
+
+set(CORE_SHARED_LIBS
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-crt-cpp.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.0unstable.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.1.0.0.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-auth.1.0.0.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-auth.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-mqtt.1.0.0.dylib
+    ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-mqtt.dylib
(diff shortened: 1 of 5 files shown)


package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
```diff
@@ -88,59 +87,77 @@ function(ly_declare_aws_library)
...
+            AWSCrt
         LIB_FILE 
-            aws-c-event-stream
+            aws-crt-cpp
         BUILD_DEPENDENCIES
-            ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-event-stream${CMAKE_SHARED_LIBRARY_SUFFIX}
-            ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-event-stream.0unstable${CMAKE_SHARED_LIBRARY_SUFFIX}
-            ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-event-stream.1.0.0${CMAKE_SHARED_LIBRARY_SUFFIX}
-            3rdParty::AWSNativeSDK::Checksums
+            ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.0unstable.dylib
```
junbo75, line 116: Nit: Is it possible to use ${CMAKE_SHARED_LIBRARY_SUFFIX} instead of the hardcoded .dylib here?

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
```diff
@@ -88,59 +87,77 @@ function(ly_declare_aws_library)
...
-#### EventStream ####
-if(LY_MONOLITHIC_GAME)
-    ly_declare_aws_library(
-        NAME 
-            EventStream
-        LIB_FILE 
-            aws-c-event-stream
-        BUILD_DEPENDENCIES 
-            3rdParty::AWSNativeSDK::Checksums
+            ${AWSNATIVE_SDK_LIB_PATH}/libaws-c-s3.a
```
junbo75, line 98: Nit: Is .a the static library? Is there any Cmake parameters (might be something like ${CMAKE_STTATIC_LIBRARY_SUFFIX}) we can use here instead of hardcoded the suffix?

junbo75 (approved):
LG

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
onecent1101: no, we should not, those variables are used if there is only one FindAWSNativeSDK.cmake.template, so suffix can be replaced properly for different platforms.
But we have different FindAWSNativeSDK for each platform, and they are not template anymore. So we should hardcode, like mac uses .dylib and .a, windows uses .dll and .lib, etc.

lawsonamzn (approved):
Looks good to me, but I'm curious about the openssl curl thing.  How are we getting openssl, and how are we getting curl, in this case?

onecent1101:
> Looks good to me, but I'm curious about the openssl curl thing. How are we getting openssl, and how are we getting curl, in this case?

it is using libs coming with xcode
for example 
```
-- Found ZLIB: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libz.tbd (found version "1.2.11")
--   Zlib library: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libz.tbd
-- Encryption: CommonCrypto
-- Http client: Curl
-- Found CURL: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libcurl.tbd (found version "7.64.1")
--   Curl include directory: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/include
--   Curl library: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libcurl.tbd
```

package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Mac
```diff
@@ -24,9 +24,8 @@ endif()
 # AWS Compile Definitions
 set(AWSNATIVESDK_COMPILE_DEFINITIONS AWS_CUSTOM_MEMORY_MANAGEMENT PLATFORM_SUPPORTS_AWS_NATIVE_SDK)
 
-set(AWSNATIVESDK_BUILD_DEPENDENCIES
-    curl # The one bundled with the aws sdk in 3rdParty doesn't use the right openssl
-)
+find_library(SECURITY_FRAMEWORK Security)
+set(AWSNATIVESDK_BUILD_DEPENDENCIES ${SECURITY_FRAMEWORK} curl)
```
lmbr-pip, line 28: nit: seems odd to lose the comment about why we explicitly define curl

lmbr-pip (approved):

PR #48, opened 2021-09-24
Update ZLIBs minimum ios, android and osx req level

lawsonamzn:
This updates the toolchain files and adds one for Mac that can be shared between different build files to ensure that everything has the correct compile flags (ie, fPIC) and also has the correct min-supported-level for the OS.  These values are the same as is expected for the O3DE engine itself.

It also updates ZLIB to use these new toolchain file(s) and cleans up the zlib build script somewhat.  Also, it adds a print message to the FindZLIB used by 3rdParties (not the one used by o3de) to indicate its 'working', as debugging this is very difficult otherwise.


package-system/zlib/FindZLIB_compat_windows.cmake
```diff
@@ -28,4 +28,6 @@ set(ZLIB_PATCH_VERSION "11")
 
 add_library(ZLIB::ZLIB INTERFACE IMPORTED GLOBAL)
 set_target_properties(ZLIB::ZLIB PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${ZLIB_INCLUDE_DIRS}")
-target_link_libraries(ZLIB::ZLIB INTERFACE ${ZLIB_LIBRARIES})
+set_target_properties(ZLIB::ZLIB PROPERTIES IMPORTED_LOCATION "${ZLIB_LIBRARIES}")
```
amznestebanpapp, line 31: looks like the only difference between these two "compat" files is the name of the lib. Instead of duplicating so much, you can handle it as a "trait" and just have these files have the set for `ZLIB_LIBRARIES`

If these Find files are entry points, they can just include a common file that sets up the rest and just have the set for `ZLIB_LIBRARIES`

package-system/zlib/Findzlib.cmake
```diff
@@ -15,30 +15,21 @@ endif()
...
+if(${PAL_PLATFORM_NAME} STREQUAL "Windows")
     set(zlib_LIBRARY_DEBUG   ${zlib_LIBS_DIR}/zlibstaticd.lib)
     set(zlib_LIBRARY_RELEASE ${zlib_LIBS_DIR}/zlibstatic.lib)
-elseif(${PAL_PLATFORM_NAME} STREQUAL "Mac")
-    set(zlib_LIBRARY_DEBUG   ${zlib_LIBS_DIR}/libz.a)
-    set(zlib_LIBRARY_RELEASE ${zlib_LIBS_DIR}/libz.a)
-elseif(${PAL_PLATFORM_NAME} STREQUAL "iOS")
+else()
     set(zlib_LIBRARY_DEBUG   ${zlib_LIBS_DIR}/libz.a)
     set(zlib_LIBRARY_RELEASE ${zlib_LIBS_DIR}/libz.a)
```
amznestebanpapp, line 23: is it intentional that both debug and release are the same library?, if so, there is no need to pass them in debug, you can just pass the `IMPORTED_LOCATION` one. You can have something like:
```
set_target_properties(${TARGET_WITH_NAMESPACE}  
    PROPERTIES
        IMPORTED_LOCATION       "${zlib_LIBRARY_RELEASE}"
)
if(zlib_LIBRARY_DEBUG)
    set_target_properties(${TARGET_WITH_NAMESPACE}  
        PROPERTIES
            IMPORTED_LOCATION_DEBUG "${zlib_LIBRARY_DEBUG}"
    )
endif()
```

package-system/zlib/build_zlib_android.cmd
```diff
@@ -5,14 +5,17 @@
...
+cmake -S temp/src -B temp/build -G Ninja ^
+    -DCMAKE_BUILD_TYPE=Release ^
+    -DCMAKE_CXX_STANDARD=17 ^
+    -DCMAKE_TOOLCHAIN_FILE=../../../../Scripts/cmake/Platform/Android/Toolchain_android.cmake ^
+    -DBUILD_SHARED_LIBS=OFF ^
+    -DSKIP_INSTALL_FILES=YES
 @if %errorlevel% NEQ 0 ( exit /b 1 )
+
 cmake --build temp/build --target zlibstatic -j 8
 @if %errorlevel% NEQ 0 ( exit /b 1 )
```
amznestebanpapp, line 19: this builds `Release` only. Are we ok with not making debug versions? 
I am ok (and prefer it) if that is the direction we want to take 3rdparty (that all are compiled in release). For Windows we would have to figure out the iterator debug level flags (and potentially other things), but for the time being we could provide the debug one.

package-system/zlib/build_zlib_mac.sh
```diff
@@ -6,15 +6,12 @@
...
-    echo "Error generating build"
-    exit 1
-fi
+# note that toolchain path is relative to the source path (-S) not to the folder this script lives in.
+cmake -S temp/src -B temp/build -G Xcode \
+    -DCMAKE_CXX_STANDARD=17 \
+    -DSKIP_INSTALL_FILES=YES \
+    -DCMAKE_TOOLCHAIN_FILE=../../../../Scripts/cmake/Platform/Mac/Toolchain_mac.cmake || exit 1
+
+cmake --build temp/build --target zlibstatic --parallel --config Release -j 8 || exit 1
```
amznestebanpapp, line 15: the same parameter is being passed: `--parallel` and `-j` are both the same: https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-tool-mode

Note that the native option for Ninja and XCode is to use all cores, so in those cases, it doesnt need to be passed.

package-system/zlib/build_zlib_linux.sh
```diff
@@ -6,14 +6,11 @@
...
-cmake -S temp/src -B temp/build -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DCMAKE_C_FLAGS=-fPIC -DSKIP_INSTALL_FILES=YES
-if [ $? -ne 0 ]; then
-    echo "Error generating build"
-    exit 1
-fi
+cmake -S temp/src -B temp/build \
+     -DCMAKE_BUILD_TYPE=Release \
+     -DCMAKE_CXX_STANDARD=17 \
+     -DCMAKE_C_FLAGS=-fPIC \
+     -DSKIP_INSTALL_FILES=YES || exit 1
```
amznestebanpapp, line 13: the default generator for Linux is `Unix Makefiles`, I suggest using Ninja instead (which we already use for Android) 
For this lib may not be a big difference, but if others are going to copy this pattern and build a big library, it will.

amznestebanpapp (changes requested):

package-system/zlib/build_zlib_android.cmd
lawsonamzn: Yes, we don't really benefit from debug zlib.  I might even remove it from windows.  We are't in the business of maintaining or debugging zlib.  And the perf of debug is slower than release.  Only reason I provide it in windows is as you say, things like iterator debug level flags.  But in this case, its a pure C library - does that even matter?

package-system/zlib/Findzlib.cmake
lawsonamzn: yes, I'll update this

package-system/zlib/FindZLIB_compat_windows.cmake
lawsonamzn: This is actually the entry point.   And I'm biasing to simplest possible file structure most readable.  There's no reason to include a common file here becuase each platform only gets their own file.  E.g. on mac you only get the one FindFile and its perfect for mac.  Adding a more complicated include heirarchy is possible but what does it buy?

package-system/zlib/build_zlib_mac.sh
lawsonamzn: will fix

package-system/zlib/build_zlib_linux.sh
lawsonamzn: Ok, wlil switch to ninja.  I take it if you dont _have_ ninja, CMAKE outputs something useful.

package-system/zlib/FindZLIB_compat_windows.cmake
lawsonamzn: The other thing to consider is that zlib being special means I'm probably not going to recommend people copy this.

What I'd much prefer is if we made a canonical version of say, libtiff  and made it behave like the official cmake one, ie, https://cmake.org/cmake/help/latest/module/FindTIFF.html

This means naming our 3p "TIFF" and not "tiff" and doing things in it like 
```
if (some variable that exists if we're within the O3DE ecosystem)
   (use o3de functions and make 3rdParty::TIFF
else()
  (make TIFF::TIFF and make it look like the output from default FindTIFF shipped with cmake
```

which woudl mean only 1 "find file" instead of having to make a 'compat".  unfortunatly the standard is ZLIB and we called ours zlib.

lawsonamzn: Which makes me think, what if I went ahead and updated this now for zlib.  Ie, changed o3de to aslk for 3rdParty::ZLIB instead of 3rdParty::zlib and then tried this?  Its worth a shot.

package-system/zlib/build_zlib_android.cmd
lumberyard-employee-dm: We only need a separate debug build of a library on Windows, if the library is using the MSVC C++ Standard library and the MSVC C++ Runtime(MSVCRTD)

package-system/zlib/build_zlib_linux.sh
lumberyard-employee-dm: Yes CMake will output that it can't find ninja during generation.

package-system/zlib/build_zlib_android.cmd
lawsonamzn: OK, so if its just a staticlib, and its pure-c it doesn't sound like we need to ship debug?

package-system/zlib/FindZLIB_compat_windows.cmake
amznestebanpapp: > Adding a more complicated include heirarchy is possible but what does it buy?

Mostly maintenance of these files, if that is a concern, I am ok dropping this. I know we are trying to simplify these scripts, so this was a potential simplification

package-system/zlib/build_zlib_android.cmd
amznestebanpapp: I think you can do it for everything except MSVC-platforms (e.g. Windows). Last time I tried, we need all 3rdparty and o3de to set `_ITERATOR_DEBUG_LEVEL` to `0` for all configurations to be able to link debug vs release. AFAIK this also applies to libs.

lawsonamzn: I see.  I should set that as a define.  HOwever, looking at the docs it seems like it only applies to C++, so C-only libraries would not be affected?

amznestebanpapp: I think it also applies to C libraries, I think is just a compiler default for `CL`
it can be passed with `-DCMAKE_CXX_FLAGS` 
But I dont think you will be able to do it, for it to work, all 3rdParty needs to either set `_ITERATOR_DEBUG_LEVEL =0` in debug or link against release. And o3de need to also set `_ITERATOR_DEBUG_LEVEL=0` in debug

Scripts/cmake/Platform/iOS/Toolchain_ios.cmake
```diff
@@ -18,3 +18,7 @@ set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
 
 set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE NO)
 
+set(CMAKE_XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET "14.0" CACHE STRING "The minimum IOS Version to support" FORCE)
```
rgba16f, line 21: This should be iOS 13.0. 

Updating all of O3DE to require iOS14 is possible, but would need wider discussion.

rgba16f: I guess the min iOS revision was updated to 14 and I didn't notice, please ignore this comment :)

amzn-sj: The Toolchain_ios.cmake for the engine was also set to iOS 14 last week. I can change it back but what are the issues with setting it to 14? Since iOS 15 came out recently, I doubt many devices would still be on iOS 13.

package-system/zlib/build_zlib_android.cmd
lawsonamzn: I'm currently testing this to make sure this works on windows debug and release.  CRT DEBUG is no longer active by default in any configuraiton, and iterators only matter if using C++.  ZLIB is compiled using pure C (it doens't even go looking for the c++ compiler or toolchain).

lumberyard-employee-dm (approved):

package-system/zlib/build_config.json
```diff
@@ -3,16 +3,15 @@
     "git_tag":"v1.2.11",
     "git_commit":"cacf7f1d4e3d44d871b605da3b647f07d718623f",
     "package_name":"zlib",
-    "package_version":"1.2.11-rev2",
+    "package_version":"1.2.11-rev5",
```
AMZN-alexpete, line 6: Should this be rev3? The PackageName is zlib-1.2.11-rev3

package-system/zlib/PackageInfo.json
```diff
@@ -1,5 +1,5 @@
 {
-    "PackageName" : "zlib-1.2.11-rev2-multiplatform",
+    "PackageName" : "zlib-1.2.11-rev3-multiplatform",
```
AMZN-alexpete, line 2: Should the package name be updated to rev5 to reflect the package_version?

lawsonamzn: This should be fixed now, I'll go double check

lawsonamzn: this file should now be gone.

AMZN-alexpete (approved):

amznestebanpapp (approved):

package-system/zlib/build_zlib_linux.sh
lawsonamzn: Done!  Didn't publish a new package, but did rerun the build script.

package-system/zlib/build_zlib_android.cmd
lawsonamzn: It works fine against windows debug

PR #49, opened 2021-09-24
[ATOM-16524] MCPP should not add new line

galibzon:
characters at the end of comment lines

This version will be "mcpp-2.7.2_az.3-rev1-<platform>"

Previously comment lines like this:
---------------------------
// Line 1
// Line 2
// Line 3
---------------------------

Would appear, after preprocessing, as
---------------------------
// Line 1

// Line 2

// Line 3

---------------------------

With this fix the number of lines is preserved per the original.

### DIFF
--- a/package-system/mcpp/.gitattributes
+++ b/package-system/mcpp/.gitattributes
@@ -1,1 +1,2 @@
 mcpp-2.7.2-az.2.patch text eol=lf
+mcpp_2.7.2-az.3.patch text eol=lf
--- a/package-system/mcpp/mcpp_2.7.2-az.3.patch
+++ b/package-system/mcpp/mcpp_2.7.2-az.3.patch
@@ -1,0 +1,897 @@
+diff --git a/config/config.guess b/config/config.guess
+index 396482d..e3ef63f 100644
+--- a/config/config.guess
++++ b/config/config.guess
+@@ -1,10 +1,9 @@
+ #! /bin/sh
+ # Attempt to guess a canonical system name.
+ #   Copyright (C) 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999,
+-#   2000, 2001, 2002, 2003, 2004, 2005, 2006 Free Software Foundation,
+-#   Inc.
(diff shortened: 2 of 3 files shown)


package-system/mcpp/mcpp_2.7.2-az.3.patch
```diff
@@ -0,0 +1,897 @@
+diff --git a/config/config.guess b/config/config.guess
+index 396482d..e3ef63f 100644
+--- a/config/config.guess
++++ b/config/config.guess
```
lawsonamzn, line 4: Is there any way to not modify config.guess?  Its license is GPL, and modifying it causes licensing concerns.
Is this maybe just an update of it?

package-system/mcpp/.gitattributes
```diff
@@ -1 +1,2 @@
-mcpp-2.7.2-az.2.patch text eol=lf
\ No newline at end of file
+mcpp-2.7.2-az.2.patch text eol=lf
+mcpp_2.7.2-az.3.patch text eol=lf
```
lawsonamzn, line 2: please add a newline at the end of this file.

package-system/mcpp/mcpp_2.7.2-az.3.patch
lawsonamzn: did we maybe pull in a new version of mcpp or something and get a bunch of junk added?

galibzon: This patch contains the same content of the original patch, + my new code in 'support.c'

package-system/mcpp/.gitattributes
galibzon: done.

package-system/mcpp/mcpp_2.7.2-az.3.patch
jeremyong-az: Why do changes to `config.guess` show up in the diff then, given that your changes only purportedly change `support.c`?

galibzon:
Closing the pull request. The MCPP folders need deep cleaning before introducing my changes to "support.c".

PR #51, opened 2021-09-27
MCPP should not add new line characters at the end of comment lines

galibzon:
Modified 'src/support.c' so comment lines do not get a new line
character appended to them. This helps keep the line count of the input
file to match before and after pre-processing.

This marks new version of the MCPP package as:
mcpp-2.7.2_az.2-rev1-<platform>


package-system/mcpp/mcpp_2.7.2_az.patch
```diff
@@ -242,15 +242,68 @@ index c57eaef..909be89 100644
...
+                 if (keep_comments) {
+-                    sp -= 2;
+-                    while (*sp != '\n')     /* Until end of line    */
+-                        mcpp_fputc( *sp++, OUT);
++                    const int numCharactersBeforeComment = (tp - temp);
++                    if (numCharactersBeforeComment == NumLeadingSpaces)
++                    {
++                        // O3DE: It is safe to print the comments
++                        sp -= 2;
++                        while (*sp != '\n')     /* Until end of line    */
```
jeremyong-az, line 299: Silly question but, why doesn't changing just this line to `*(sp + 1) != '\n'` not work to address the issue described in the description?

galibzon: It's hard to do code review over the diff of a patch...
Maybe in the future we should move MCPP to GIT. Will need to follow up with the original author.

To see what's going on you'd have to see the original 'src/support.c' and that's why I added an //O3DE next to the lines I'm inserting. The original code is a massive pure 'c' state machine that is brittle and prone to bugs to any small change in pointer arithmetic. It took me a while to understand that state machine and believe me when I say that innocent or trivial changes to the pointer math they do always has collateral damage somewhere else ;-)

With this PR the goal is accomplished, not worth investing more time than necessary.

galibzon: Here is the original snippet:
```
                if (keep_comments) {
                    sp -= 2;
                    while (*sp != '\n')     /* Until end of line    */
                        mcpp_fputc( *sp++, OUT);
                    mcpp_fputc('\n', OUT);
                }
```

jeremyong-az: I actually downloaded the original package when I was doing the review because it's really hard to review a diff/patch as you say. The original code looks like this:

```c
                if (keep_comments) {
                    sp -= 2;
                    while (*sp != '\n')     /* Until end of line    */
                        mcpp_fputc( *sp++, OUT);
                }
                goto  end_line;
```
It seems like all we want to do is strip a single `\n` from the output stream, so something like this right?
```c
                if (keep_comments) {
                    sp -= 2;
                    while (*(sp + 1) != '\n')     /* Until one character before the end of line    */
                        mcpp_fputc( *sp++, OUT);
                    ++sp; // Advance to original position
                }
                goto  end_line;
```

jeremyong-az: Is the issue specifically things like carriage return characters maybe?

galibzon: No. You see it says:
`goto  end_line;`
At the start of 'goto end_line' there's another:

jeremyong-az (approved):

santorac (approved):

spham-amzn (approved):

package_build_list_host_darwin.json
```diff
@@ -79,7 +79,7 @@
         "mikkelsen-1.0.0.4-ios": "package-system/mikkelsen-ios",
         "poly2tri-7f0487a-rev1-mac": "package-system/poly2tri-mac",
         "v-hacd-2.3-1a49edf-rev1-mac": "package-system/v-hacd-mac",
-        "mcpp-2.7.2_az.1-rev1-mac": "package-system/mcpp-mac",
+        "mcpp-2.7.2_az.2-rev1-mac": "package-system/mcpp-mac",
```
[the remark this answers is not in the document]
galibzon: It's a style in some packages when there are errors when uploading package and you need to resubmit another one. You only change the revision but not the version.

PR #54, opened 2021-10-07
Enable package dependencies and package tests

nick-l-o3de:
Enables the ability for packages to depend on each other (auto-gets the packages it depends on). 
Enables the ability for package scripts from pull_and_build_from_git.py to have automatic self tests.
Updates the TIFF package to fully utilize this feature.


Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,134 @@
...
+        
+            return hash_result
+
+        # make sure a package with that name is not already present:
+        server_urls = os.environ.get("LY_PACKAGE_SERVER_URLS", default = "")
+
+        if not server_urls:
+            print(f"Server url list is empty - please set LY_PACKAGE_SERVER_URLS env var to semicolon-seperated list of urls to check")
+            print(f"Using default URL for convenience: https://d2c171ws20a1rv.cloudfront.net")
+            server_urls = "https://d2c171ws20a1rv.cloudfront.net"
```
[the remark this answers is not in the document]
lawsonamzn: Good idea here.

Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,134 @@
...
+            finds the package, if possible, and downloads and unpacks it to a given folder.
+            Note that this essentially mimics the server urls protocol used by cmake on the client
+            side and thus will also check buckets if s3:// urls are given
+            and will stop at the first one it finds.
+
+            Assumes that LY_PACKAGE_SERVER_URLS is set in the environment
+            Assumes that, if necessary, LY_AWS_PROFILE is set in the environment (if using s3 buckets)
+
+            returns True if it succeeded, False otherwise.
+         '''
```
sptramer, line 32:
```suggestion
        '''Given a semicolon-separated public server URL list (semicolon-seperated), 
            attempts to find the package. If found, downloads and unpacks to the target_folder location.
            Only the first found package is downloaded, and then the search stops. If the checksum of
            the downloaded file doesn't match the checksum in the O3DE dependency list, the package
            isn't unpacked on the filesystem and the download is deleted.
        
            This method supports all URI types handled by the O3DE package system, including S3 URIs.
            
            PRECONDITIONS:
            * LY_PACKAGE_SERVER_URLS must be set in the environment
            * If using S3 URIs, LY_AWS_PROFILE must be set in the environment
            
            Returns True if successful, False otherwise.
         '''
```

Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,134 @@
...
+                continue
+            full_package_url = package_server + "/" + package_file_name
+            print(f"    - attempting '{full_package_url}' ...")
+
+            try:
+                if package_server.startswith("s3://"):
+                    # we don't want to use temp buckets for package dependencies
+                    # you should depend on production (CDN) resources that have been verified!
+                    print(f"        - Note:  Ignoring {package_server} - S3 buckets should not be used for package dependencies.")
+                    continue
```
sptramer, line 81: This is on the user, not on the software to detect. There are legitimate long-term S3 stores that could be used for a dependency system for consumers creating their own 3P packages.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -90,6 +99,31 @@
...
+    - note that we don't check recursively - you must name your recursive deps!
+    - The packages must be on a public CDN or locally tested with FILE:// - it uses env var 
+      "LY_PACKAGE_SERVER_URLS" which can be a semicolon seperated list of places to try.
+    - The packages unzip path + subfolder is added to CMAKE_MODULE_PATH if you use cmake commands.
+    - Otherwise you can use DOWNLOADED_PACKAGE_FOLDERS env var in your custom script and set
+    - CMAKE_MODULE_PATH to be that value, yourself.
+    - The subfolder can be empty, in which case the root of the package will be used.
+
+Note about environment variables:
+When custom commands are issued (build, install, and test), the following environment variables will be set
```
sptramer, line 117:
```suggestion
When custom commands are issued (build, install, and test), the following environment variables will be set for the process:
```

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -593,30 +649,51 @@ def build_and_install_cmake(self):
...
-        custom_install_env = os.environ.copy()
-        custom_install_env['TARGET_INSTALL_ROOT'] = str(self.build_install_folder.resolve())
+       
         for custom_install_cmd in custom_install_cmds:
             call_result = subprocess.run(custom_install_cmd,
                                          shell=True,
                                          capture_output=False,
                                          cwd=str(self.base_folder),
-                                         env=custom_install_env)
+                                         env=self.create_custom_env())
```
sptramer, line 696: Does `self.create_custom_env()` need to be repeated here, or are different 3P package configs allowed for `build` and `install`?

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -765,6 +856,8 @@ def execute(self):
         # Generate the Find*.cmake file
         self.generate_cmake()
 
+        self.test_package()
```
sptramer, line 859: :+1::+1::+1::+1:

package-system/tiff/build_tiff_android.cmd
```diff
@@ -5,14 +5,28 @@
...
 @rem #
 
-@IF "%LY_ANDROID_NDK_ROOT%" == "" (
-    @echo Need to set the varible LY_ANDROID_NDK_ROOT to the location of your NDK install!
-    @exit /b 1
-)
+@rem # note that we explicitly turn off the compilation of all features that rely on 3rd Party Libraries
+@rem # except the ones we want.  This prevents the cmake build system from automatically finding things
+@rem # if they happen to be installed locally, which we don't want.
+cmake -S temp/src -B temp/build -G Ninja ^
```
sptramer, line 11: Problem here that I've seen with double-linking static libs on Android and other platforms - if a build product links its own libz somewhere (very likely) then this could result in build incompatibilities and problems resolving symbols in a monolithic binary. Does not need to be addressed here but we may require some advanced user documentation on which 3P libraries are built by O3DE, what their default configurations are, and where to make modifications to builds.

Maybe things have changed but I'm curious if this remains an issue on static-link platforms under certain compilers.

sptramer (changes requested):
Minor changes for cleaner API docs and some questions about code changes or possible future issues.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -593,30 +649,51 @@ def build_and_install_cmake(self):
...
+            package_folder_list = ""
+            for package_name, _, subfoldername in self.package_info.depends_on_packages:
+                if package_folder_list:
+                    package_folder_list += ";"
+                # note that it must be an abs path for cmake to accept it:
+                # The package declaration may specify a subfolder name but the path library will
+                # behave correctly if you slash operator with an empty (by ignoring it)
+                package_folder_list += str( (self.base_temp_folder / package_name / subfoldername).resolve().absolute())
+            package_folder_list = package_folder_list.replace('\\', '/')
+            custom_env['DOWNLOADED_PACKAGE_FOLDERS'] = package_folder_list
```
spham-amzn, line 667: minor

string.join can be used so you don't need an extra condition to add ';'

```suggestion
            package_folder_list = []
            for package_name, _, subfoldername in self.package_info.depends_on_packages:
                # note that it must be an abs path for cmake to accept it:
                # The package declaration may specify a subfolder name but the path library will
                # behave correctly if you slash operator with an empty (by ignoring it)
                package_folder_list += str( (self.base_temp_folder / package_name / subfoldername).resolve().absolute())
            package_folder_list = package_folder_list.replace('\\', '/')
            custom_env['DOWNLOADED_PACKAGE_FOLDERS'] = ';'.join(package_folder_list).replace('\\', '/')
```

Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,134 @@
...
+                print("    - Warning: Hash of package does not match - will not use it")
+                os.remove(package_download_name)
+                continue
+
+            # hash matched.  Unpack and return!
+            if not os.path.exists(str(package_unpack_folder)):
+                os.makedirs(str(package_unpack_folder))
+            with tarfile.open(package_download_name) as archive_file:
+                print("    - unpacking package...")
+                archive_file.extractall(package_unpack_folder)
```
spham-amzn, line 105: Consider wrapping this with a try/except block in case of corrupted tar files

Scripts/extras/pull_and_build_from_git.py
lawsonamzn: no, its the same config.  I could remove the recreation of it.  The only danger is if something somehow modifies the env, so you get a clean env, but I'm not sure that's possible.

package-system/tiff/build_tiff_android.cmd
lawsonamzn: Yes.  Of course, this new zlib package is actually compatible with those other libraries.  So for example, I've already tested it but tifflib, as well as openexr, openimageio, and python can all use the O3DE Zlib as a drop in replacement.  
Not to be addressed in this, of course.  But going forward if we do have a 3p package that satisifies a dep, I'd prefer if we compile using that instead of a copy.

Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,134 @@
...
+            print(f"    - attempting '{full_package_url}' ...")
+
+            try:
+                if package_server.startswith("s3://"):
+                    # we don't want to use temp buckets for package dependencies
+                    # you should depend on production (CDN) resources that have been verified!
+                    print(f"        - Note:  Ignoring {package_server} - S3 buckets should not be used for package dependencies.")
+                    continue
+                tls_context = ssl.create_default_context(cafile=certifi.where())
+                with urllib.request.urlopen(url=full_package_url, context = tls_context) as server_response:
```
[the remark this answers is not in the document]
lawsonamzn: yes, its how I was testing without uploading.

lawsonamzn:
I've updated it to fully support s3 bucketing.  I also retested building it on windows, mac.  I also reversed the order and also tested it without boto installed and without access to the s3 bucket to ensure that the appropriate errors occur.

spham-amzn (approved):

Scripts/extras/package_downloader.py
sptramer: If it can't plug to the existing S3 package infra yet, good enough reason.

sptramer (approved):

Scripts/extras/package_downloader.py
```diff
@@ -0,0 +1,177 @@
...
+                    with urllib.request.urlopen(url=full_package_url, context = tls_context) as server_response:
+                        print("    - Downloading package...")
+                        file_data = server_response.read()
+                        with open(package_download_name, "wb") as save_package:
+                            save_package.write(file_data)
+            except Exception as e:
+                # note that anything that causes this to fail should result in trying the next one.
+                print(f"        - Unable to get package from this server: {e}")
+                continue # try the next URL, if any...
+
```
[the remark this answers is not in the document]
lawsonamzn: What would you suggest we do?  We want to ignore all exceptions a server might throw, including all file based (extraction / untar), http, cert, file, signing, etc, and move onto the next server

lumberyard-employee-dm (approved):

package-system/tiff/build_config.json
```diff
@@ -2,60 +2,91 @@
...
                 ],
                 "custom_install_cmd": [
                     "install_tiff_windows.cmd"
+                ],
+                "custom_test_cmd" : [
+                    "test_tiff_windows.cmd"
                 ]
             },
             "Android":{
+                "package_version":"4.2.0.15-rev4",
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Well that is useful to know going forward

PR #59, opened 2021-10-19
Added support for being able to dynamically reference the python executable that launched the build script from the build_config.json

cgalvan:
When using the `custom_build_cmd` in a `build_config.json`, it would be useful to be able to invoke the python executable that ran the build script (e.g. the python runtime from the O3DE engine), since the `python` on the system PATH might not be the one you would want to invoke.

This would allow you to do something like this:

```
"custom_build_cmd": [
    "{python} foo.py",
    "./someOtherScript.sh"
],
```

The use-case that I found for this is I'm working on a build script for Qt on Mac where I'd like to execute a python script using our Python3 executable, because the default python on the system PATH is still Python 2.7.

Signed-off-by: Chris Galvan <chgalvan@amazon.com>

### DIFF
--- a/Scripts/extras/pull_and_build_from_git.py
+++ b/Scripts/extras/pull_and_build_from_git.py
@@ -674,8 +674,9 @@ class BuildInfo(object):
         env_to_use = self.create_custom_env()
         custom_build_cmds = self.platform_config.get('custom_build_cmd', [])
         for custom_build_cmd in custom_build_cmds:
-
-            call_result = subprocess.run(custom_build_cmd,
+            # Support the user specifying {python} in the custom_build_cmd to invoke
+            # the Python executable that launched this build script
+            call_result = subprocess.run(custom_build_cmd.format(python=sys.executable),
                                          shell=True,
                                          capture_output=False,
                                          cwd=str(self.base_folder),


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -674,8 +674,9 @@ def build_and_install_custom(self):
         env_to_use = self.create_custom_env()
         custom_build_cmds = self.platform_config.get('custom_build_cmd', [])
         for custom_build_cmd in custom_build_cmds:
-
-            call_result = subprocess.run(custom_build_cmd,
+            # Support the user specifying {python} in the custom_build_cmd to invoke
+            # the Python executable that launched this build script
+            call_result = subprocess.run(custom_build_cmd.format(python=sys.executable),
```
lumberyard-employee-dm, line 679: This seems a little too custom for a function that is meant to invoke an arbitrary command.
There is no guarantee that the builder of the 3rdparty library has the o3de repo on their machine with the o3de python downloaded.

We shouldn't require our python in order to build 3rdParty libraries using the system.
Now having a system requirement on `python3` seems fine, but it doesn't have to be the one that the 3rdParty system downloads.

cgalvan: I thought of it because I noticed our docs show using the o3de python, but you're right it should be able to work with any python.

I was having an issue because I was getting some different issues when having multiple different pythons on the PATH, so I thought it would be helpful to be able to dynamically use the python executable that had launched the build script, but if we can make having `python3` on the PATH a requirement then I am happy with that and can drop this PR and use that instead.

amznestebanpapp: I am a bit confused about the comment, maybe I am missing something. This is not requiring o3de's python, right? its just using the python that called this script to execute the command. 

Does this still work if it is not a py command being executed? (e.g. a batch)

lumberyard-employee-dm: This behavior still seems strange to be a special case just for python.
It doesn't exist for invoking `cmake`, `clang`, `bash`, etc... for a custom build command.

lawsonamzn: While I agree its a special case, because it doesn't exist for bash, clang, cmake, etc, I think its okay because of context.  The context here is that the actual build_settings.json file that will contain this "command to run" tag  is being parsed by pull_and_build_from_git.py, which is in fact a python script, and its not unreasonable want to run any sub-python shells using the same python that was used as the entry point to the script.  

I'd actually go so far as to add the %PYTHON% env var ($PYTHON) to the env of the custom commands that are using bash so that non-python (bash, sh, etc) can use the right python too.

lawsonamzn (approved):
I'm okay with this. But ocnsider updating the helptext near the top of the file with this new var?

Also, is PYTHON_EXECUTABLE safe to use?  We'd want to set a varialbe name that had no impact if unused by the script being called.  PYTHON_EXECUTABLE is probably safe.  PYTHONEXECUTABLE would definitely not be safe.

cgalvan:
> I'm okay with this. But ocnsider updating the helptext near the top of the file with this new var?
> 
> Also, is PYTHON_EXECUTABLE safe to use? We'd want to set a varialbe name that had no impact if unused by the script being called. PYTHON_EXECUTABLE is probably safe. PYTHONEXECUTABLE would definitely not be safe.

Good call out!  Will replace with `PYTHON_BINARY` instead and will add help text.

PR #63, opened 2021-11-02
OpenMesh 3P Source Scripts for support for Linux, Mac, and Windows

spham-amzn:


lawsonamzn (commented):
So some changes I'd reccomend here
* if you've already split the find files out for each platform, no point in using the $${} version, may as well de-templatize them and just use the find source field instead of the find template field.  Makes it easier to read and work with
* please use imported_location and so on, instead of link_libraries for the actual imported libraries.  
* no point in using debug libraries on mac or linux.   If its a pure C library you dont even have to use debug libraries on windows.
* If the various find files are identical between unixlikes consider using the same one for all fo them?

package-system/OpenMesh/FindOpenMesh.cmake.Linux
```diff
@@ -0,0 +1,34 @@
...
+
+set($${LIB_NAME}_LIBRARY_DEBUG   $${$${LIB_NAME}_LIBS_DIR}/libOpenMeshCored.a)
+set($${LIB_NAME}_LIBRARY_RELEASE $${$${LIB_NAME}_LIBS_DIR}/libOpenMeshCore.a)
+
+set($${LIB_NAME}_LIBRARY
+    "$$<$$<CONFIG:profile>:$${$${LIB_NAME}_LIBRARY_RELEASE}>"
+    "$$<$$<CONFIG:release>:$${$${LIB_NAME}_LIBRARY_RELEASE}>"
+    "$$<$$<CONFIG:debug>:$${$${LIB_NAME}_LIBRARY_DEBUG}>")
+
+add_library($${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
```
lawsonamzn, line 27: this is a static library, not an interface, so declare it as STATIC IMPORTED GLOBAL, instead of INTERFACE IMPORTED GLOBAL  (same with the rest of the find fiiles).  And supply an IMPORTED_LOCATION, not target_link_libraries
```
set_target_properties(${TARGET_WITH_NAMESPACE} PROPERTIES IMPORTED_LOCATION "${${LIB_NAME}_LIBRARY}")
```

package-system/OpenMesh/build_config.json
```diff
@@ -0,0 +1,88 @@
...
+            "cmake_generate_args_debug":[
+               "-DDISABLE_QMAKE_BUILD=ON",
+               "-DBUILD_APPS=OFF",
+               "-DOPENMESH_DOCS=OFF",
+               "-DOPENMESH_BUILD_SHARED=OFF",
+               "-DOPENMESH_BUILD_UNIT_TESTS=OFF",
+               "-DCMAKE_CXX_FLAGS=\"-fPIC -fno-aligned-allocation\"",
+               "-DCMAKE_CXX_STANDARD=17",
+               "-DCMAKE_BUILD_TYPE=Debug"
+            ],
```
lawsonamzn, line 35: You will need to also supply the mac toolchain file to make sure the right version of mac is given - see the build scripts for tiff and zlib.

spham-amzn:
>     please use imported_location and so on, instead of link_libraries for the actual imported libraries.

Trying this out does not seem to work with the generator expression we use for debug vs release on windows. 

> no point in using debug libraries on mac or linux. If its a pure C library you dont even have to use debug libraries on windows.

We still need debug libraries on windows since this is not a pure C library

lawsonamzn:
> > ```
> > please use imported_location and so on, instead of link_libraries for the actual imported libraries.
> > ```
> 
> Trying this out does not seem to work with the generator expression we use for debug vs release on windows.
> 

Sorry, it works a different way than link libraries:
```
set_target_properties(${TARGET_WITH_NAMESPACE} PROPERTIES IMPORTED_LOCATION "${${MY_NAME}_LIBRARY_RELEASE}")

#optional:  If you'd like to override the above default behavior for a particular configuration like DEBUG or PROFILE:
set_target_properties(${TARGET_WITH_NAMESPACE} PROPERTIES IMPORTED_LOCATION_DEBUG "${${MY_NAME}_LIBRARY_DEBUG}")
```

Cmake will use the imported_location by default, and then allow you to override for configuration otherwise.

lawsonamzn (approved):
looks great!

lawsonamzn (approved):
Why doesn't Xcode gen work?  its a valid generator name.  I guess unix makefiles is the default if that doesnt work?

spham-amzn:
> 
> 
> Why doesn't Xcode gen work? its a valid generator name. I guess unix makefiles is the default if that doesnt work? Also, we want staticlib, right, not shared?

OpenMesh docs for mac is very sparse:

https://www.graphics.rwth-aachen.de/media/openmesh_static/Documentations/OpenMesh-7.1-Documentation/a03923.html

But from the looks of it, it only supports make.

spham-amzn:
FYI, OpenMesh appears to use something called "ACG Buildsystem"

```
************************************************************
* ACG Buildsystem                                          *
*                                                          *
* Package : OpenMesh                                       *
* Version : 8.1                                            *
* Type    : Release                                        *
************************************************************
```

spham-amzn:
> Also, we want staticlib, right, not shared

Their CMakeLists.txt only controls static vs shared on Windows. For Mac/Linux, there is no control and both are built

package-system/OpenMesh/openmesh-o3de.patch
```diff
@@ -0,0 +1,193 @@
...
++  vectorize(_n, Scalar(0.0)); // o3de fix
+   for (ConstVertexFaceIter vf_it = this->cvf_iter(_vh); vf_it.is_valid(); ++vf_it)
+     _n += this->normal(*vf_it);
+ }
+@@ -484,7 +492,7 @@ template <class Kernel>
+ void PolyMeshT<Kernel>::
+ calc_vertex_normal_correct(VertexHandle _vh, Normal& _n) const
+ {
+-  vectorize(_n, Scalar(0));
++  vectorize(_n, Scalar(0.0)); // o3de fix
```
hultonha, line 177: suggestion: It looks like we might actually be able to remove these diffs, in OpenMesh 8.0 there was no cast, which is when we added it, but I think this was fixed in 8.1, our only change is making it `0.0` which isn't strictly required, so we could just revert these lines and keep vanilla OpenMesh

hultonha:
Tagging @pollend here too for visibility as this is closely related to https://github.com/o3de/o3de/pull/5075

package-system/OpenMesh/openmesh-o3de.patch
spham-amzn: Yeah, maybe for the next upgrade. It takes about half a day to go through the build and testing for all 3 platforms whenever I update the package, and I don't this will really affect the final binary much. But minimizing/eliminating our patches eventually is desirable

lawsonamzn (approved):

lawsonamzn:
I can merge this in if you'd like and think its ready

As for the tests, using this on the other side (ie, in o3de) involves making a PR that uses it and presumably runs all the tests.

spham-amzn:
> 
> 
> I can merge this in if you'd like and think its ready
> 
> As for the tests, using this on the other side (ie, in o3de) involves making a PR in o3de that uses it and thus presumably runs all the tests via AR

PR for that is here:

https://github.com/o3de/o3de/pull/5248

package-system/OpenMesh/openmesh-o3de.patch
hultonha: Sure thing, totally makes sense :+1:

PR #66, opened 2021-11-09
Fix errors and improvements for Physx SDK package

moraaar:
- Fixed physx build script (missing comma and misspelled library)
- Added "checked" configuration so customers can use it to look for errors in PhysX.
- Using clearer names "static" and "shared" for static and dynamic versions of the library.

Signed-off-by: moraaar <moraaar@amazon.com>


yuriy0:
This is a very important change - having the checked physx build in o3de profile builds is very helpful for tracking down bugs (and usually the o3de debug build is simply too slow). Thank you for putting this together.

Perhaps consider adding another build variant for the "debug" physx build, which enables even more validation compared to "checked", which is also useable in profile o3de builds. On windows, the "debug" variant of physx is built using the debug CRT so is not compatible with the o3de profile build. This means there would have to be two different "debug" variants of physx - one with the debug CRT, one without. To be concrete, I would suggest:
- "debug" o3de always uses "debug" physx
- "release" o3de always uses "release" physx
- "profile" o3de uses a different physx build based on some cmake flag, e.g. "PHYSX_DEBUGGING_MODE=Profile" uses the "profile" physx build, "PHYSX_DEBUGGING_MODE=Debug" uses the "debug_non_debug_crt" physx build, "PHYSX_DEBUGGING_MODE=Checked" uses the "checked" physx build

moraaar:
Patching vcpckg so it also consider the "Checked" configuration that PhysX SDK already has was trivial because the work was already done to add "Profile". But adding a fourth configuration "debug_non_debug_crt" would require to patch both vcpck and PhysX SDK itself. 

I think checked configuration activates already all the necessary validations checks in PhysX, so I suggest try that for now.

I do like your idea of defining a macro to be able to choose what configuration to use with Profile, so you could choose either Profile (default), Checked (if you want further validations) or Release (if you want to test performance without having to build in Release).

I'll change FindPhysX.cmake to check if `PHYSX_PROFILE_CONFIG_OVERRIDE` is set during cmake config, and if it does, then it uses, else it uses Profile by default.

AMZN-alexpete:
The PhysX package is already our 3rd largest package at 111MB compressed and ~1GB extracted.  How much larger will adding the checked libraries make this?  Keep in mind this is additional data that every O3DE user will have to download whether they need to debug PhysX or not.  Also, if the debug libraries are too slow to be of any use, can we find a way to not include them in the package or make the optional?

moraaar:
> The PhysX package is already our 3rd largest package at 111MB compressed and ~1GB extracted. How much larger will adding the checked libraries make this? Keep in mind this is additional data that every O3DE user will have to download whether they need to debug PhysX or not. Also, if the debug libraries are too slow to be of any use, can we find a way to not include them in the package or make the optional?

By adding checked libs the compressed side increases to 145MB (+34MB) and 1.46GB (+341MB) extracted. I think the extra size is acceptable considering the benefits this brings and how complicated would be for customers to recompile PhysX SDK to get the checked version.

Another questions would be if it's worth separate the packages even more, by static vs release or by configuration. But considering that would have to be done for all platforms (windows, mac, ios, android and linux) I rather not make the PhysX SD 3rdParty more complicated to save a few MBs to download. For now I think to keep the packages separated by platform but keeping all skews in one package is good enough.

The PhysX debug libraries are still useful, what could be too slow is for a particular customer to run theirs projects in O3DE itself in debug, but that's a separate issue.

moraaar:
I've included the macro `LY_PHYSX_PROFILE_USE_CHECKED_LIBS` which is cached by default to OFF, by doing so it will be discoverable by cmake-gui.

lawsonamzn (approved):
At some point we probably have to wean ourselves off VCPKG for the purposes of supporting non-standard platforms (Which vcpkg doesn't) but for now this seems good to me.

greerdv (approved):

PR #68, opened 2021-11-16
Upgrade astc-encoder to use SSE4 on all platforms

lawsonamzn:
Note the addition of qt-* and such to the ignore file just reduces clutter in 'git status'.  I can eliminate those if it makes people feel better.


package-system/astc-encoder/install_mac.sh
```diff
@@ -17,8 +17,8 @@ cp -f temp/src/Source/astcenc.h $INCLUDE_PATH/ || exit $?
 
 BUILD_PATH=temp/build/Source
 
-# copy static lib and executable
-cp -f $BUILD_PATH/astcenc-native $BIN_PATH/ || exit $?
-cp -f $BUILD_PATH/libastcenc-native-static.a $BIN_PATH/ || exit $?
+# copy sse2 lib and executable
```
rgba16f, line 20: comment says sse2, should be sse4?

rgba16f (approved):

package-system/astc-encoder/install_mac.sh
lawsonamzn: nice catch.  I can update this without invalidating the packages, thankfully

spham-amzn (approved):

PR #70, opened 2021-11-19
Fix Findpyside2.cmake and update package rev

spham-amzn:
Fix bad Findpyside2.cmake from the previous production push and update the rev number.
The original Findpyside2.cmake for Linux was missing the 'ly_add_target_files' which prevented the pyside libraries not to be copied over as a runtime dependency


package-system/pyside2-qt/Findpyside2.cmake.Linux
```diff
@@ -5,24 +5,28 @@
...
 set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
 if (TARGET ${TARGET_WITH_NAMESPACE})
     return()
 endif()
-     
+
 add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
-     
+
 ly_pip_install_local_package_editable(${CMAKE_CURRENT_LIST_DIR}/pyside2 pyside2)
```
NicholasVanSickle, line 16: I assume this should be removed in favor of the later call with `PYSIDE_BASE_PATH`

spham-amzn: oh, good catch, looks like a duplicate call (see line 20)

cgalvan (approved):

lawsonamzn (approved):

PR #75, opened 2021-12-22
bringing assimp over to this repository

AMZN-stankowi:
### Done

- Generated a patch with our changes to assimp
- Discovered which commit of assimp our build was based off of
- Added new logic for the install step to be driven by json, because for this library it's just copying files out of the package
- Windows build, install, test step
- Mac build, install, test step
- Linux build, install, test steps
- Generate actual packages
- Test and verify this works end to end
- 
### To be done next, in a separate pull request

- Push those packages up to the package repository
- Update O3DE to use these packages

### To be done later, not as part of this pull request

- Update to the latest version of assimp
- Create pull requests of our outstanding changes to canonical assimp
- Decide if we want to break up these patches into the individual changes, or keep them as one monolithic patch
- Expanded tests that match how O3DE consumes this library, to catch things like linker issues early.


package-system/assimp/install_assimp_linux.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Source_Subfolder": "src",
+  "Install_Paths": [
+    "include/",
+    "lib/libassimp.a",
+    "bin/assimp",
+    "bin/libassimp.so.5.0.1",
+    "bin/unit",
```
lawsonamzn, line 8: Where does this come from?  is it something that assimp makes?

lawsonamzn: I meant specifically the 'unit' executable.

AMZN-stankowi: yeah, that's the unit test executable. Previously we weren't including it in the package, but I decided to put it in because I thought it would be useful when debugging tougher problems with assimp. Running the unit test executable locally can server as a good step to verify a baseline of functionality is working correctly.

I've also set it to run in the test step, so it's run as part of this packaging work.

lawsonamzn: Definitely good to run the tests that are included in the package but an arguably more important test is one that uses the package itself.   You could save yourself a lot of worry and trouble that way since it short cuts the iteration cycle of ingesting this package in O3DE.

What I mean by a test that uses the package itself is, e.g. a separate folder in your 3p-package-source location (like a subfolder called test or something). Which contains a main.cpp and a CMakeLists.txt that declares an executable and calls find_package(ASSIMP).   As long as the _final output folder_ of the package (plus where zlib was dropped) is in the CMAKE_MODULE_PATH, it should work, and it should work identically to how its likely to work in the O3DE side, revealing any linkage errors or include problems.

AMZN-stankowi: Good idea, I've added it to the list of follow up work I want to do after getting the first part of this done.

package-system/assimp/build_assimp_mac.sh
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+cmake -S temp/src -B temp/build \
+    -DCMAKE_CXX_COMPILER=clang++ \
```
spham-amzn, line 10: Is mac using clang and linux using gcc?

AMZN-stankowi: I'm not too sure, I'm not very familiar with either platform, these two scripts are what I was able to get working after a lot of trial and error and looking at all the other 3rd party libraries.

spham-amzn: Reason I ask is because it seems you should be able to build with the same args for both mac and linux, so you can just use one script for both platforms. I just recently did the same thing for SQLite (it wasn't cmake though).

AMZN-stankowi: After I figure out this Windows failure that's occurring with the patch, I'll try the Linux script on mac then and see what I get. When I tried to use this command on Linux it couldn't find Clang, I ended up spending a lot of time on the internal wiki looking up things to try until some combination of "apt get" and that iteration of the script worked.

package-system/assimp/0001-O3DE-fixes-applied-on-top-of-376e46887780e5090af0198.patch
```diff
@@ -0,0 +1,317 @@
...
+ names are added to the output here.
+
+Signed-off-by: AMZN-stankowi <4838196+AMZN-stankowi@users.noreply.github.com>
+---
+ code/AssetLib/Assxml/AssxmlFileWriter.cpp |   7 +-
+ code/AssetLib/FBX/FBXConverter.cpp        | 142 ++++++++++++++++------
+ code/AssetLib/FBX/FBXProperties.cpp       |   2 +-
+ include/assimp/material.h                 |  12 ++
+ include/assimp/mesh.h                     |   2 +
+ 5 files changed, 125 insertions(+), 40 deletions(-)
```
spham-amzn, line 23: Why is this in the patch?

AMZN-stankowi: I just placed the Perforce copy of assimp on top of the github commit, and generated a patch. Is there a better way to do it?

spham-amzn (approved):

PR #77, opened 2022-01-05
SQLite 3P Conversion

spham-amzn:
* Update to pull from the github mirror for SQLite : https://github.com/sqlite/sqlite/tree/version-3.32.2
* Build using the source repo's make/nmake scripts
* Minor updated to pull_and_build_from_git.py to accomodate the custom license file

* Tested on Windows/Linux
* Testing on Mac currently


package-system/sqlite/build_config.json
```diff
@@ -0,0 +1,46 @@
...
+            ]
+         }
+      },
+      "Darwin":{
+         "Mac":{
+            "custom_build_cmd": [
+               "./build_linux.sh"
+            ],
+            "custom_install_cmd": [
+               "./package_linux.sh"
```
lumberyard-employee-dm, line 31: Seems like `build_unix.sh` and `package_unix.sh` is a more accurate name for these scripts.

package-system/sqlite/FindSQLite.cmake
```diff
@@ -0,0 +1,37 @@
...
+set(${MY_NAME}_LIBS_DIR ${CMAKE_CURRENT_LIST_DIR}/SQLite/lib)
+
+set(${MY_NAME}_LIBRARY_DEBUG   ${${MY_NAME}_LIBS_DIR}/debug/${CMAKE_STATIC_LIBRARY_PREFIX}sqlite3${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(${MY_NAME}_LIBRARY_RELEASE ${${MY_NAME}_LIBS_DIR}/release/${CMAKE_STATIC_LIBRARY_PREFIX}sqlite3${CMAKE_STATIC_LIBRARY_SUFFIX})
+
+# we set it to a generator expression for multi-config situations:
+set(${MY_NAME}_LIBRARY  
+    "$<$<CONFIG:profile>:${${MY_NAME}_LIBRARY_RELEASE}>"
+    "$<$<CONFIG:Release>:${${MY_NAME}_LIBRARY_RELEASE}>"
+    "$<$<CONFIG:Debug>:${${MY_NAME}_LIBRARY_DEBUG}>")
```
lumberyard-employee-dm, line 27: The `${MY_NAME}_LIBRARY` variable in the CMake Find scripts are no longer needed.
Because we specify the target type correctly to be STATIC, we can use the `IMPORTED_LOCATION` property to set the library location.

package-system/sqlite/FindSQLite.cmake
```diff
@@ -0,0 +1,37 @@
...
+
+add_library(${TARGET_WITH_NAMESPACE} STATIC IMPORTED GLOBAL)
+set_target_properties(${TARGET_WITH_NAMESPACE} PROPERTIES IMPORTED_LOCATION "${${MY_NAME}_LIBRARY_RELEASE}")
```
lumberyard-employee-dm, line 30: I don't see a real benefit of using the Debug version of sqlite3 as I doubt we need to debug it and the performance of the Release is a lot better.

Also since sqlite3 is a C library that doesn't use the MSVC runtime we can use the Release version on Windows as well.

package-system/sqlite/build_config.json
spham-amzn: ```build_unix_like.sh``` and ```package_unix_like.sh``` ? (from https://sqlite.org/src/doc/trunk/README.md )

lumberyard-employee-dm: "unix_like" is more accurate and I am good with that suggestion.

lumberyard-employee-dm (approved):

lawsonamzn (approved):

PR #78, opened 2022-01-20
Expat conversion

lawsonamzn:
Adds a new option to the pull_and_build_from_git.py script to allow packages that do not have a cmakelists in their root folder but instead in a subfolder.

Moves expat to the latest stable release and adds build scripts for it

### DIFF
--- a/package-system/expat/Findexpat.cmake
+++ b/package-system/expat/Findexpat.cmake
@@ -14,7 +14,20 @@ endif()
 # Even though expat itself exports it as lowercase expat, older cmake (and cmake's built-in targets)
 # expect uppercase.  So we define both, for backwards compat:
 
-set(EXPAT_LIBRARY ${CMAKE_CURRENT_LIST_DIR}/expat/lib/${CMAKE_STATIC_LIBRARY_PREFIX}expat${CMAKE_STATIC_LIBRARY_SUFFIX})
+if (WIN32)
+    # on windows, expat adds the nonstandard 'lib' prefix and MD, dMD suffixes for 
+    # Multighreaded Dynamic CRT and debug Multithreaded Dynamic CRT
+    # We don't use the debug version since its a pure C library with no C++ and thus will
+    # not have an ITERATOR_DEBUG_LEVEL conflict
+    set(PREFIX_TO_USE "lib")
+    set(SUFFIX_TO_USE "MD.lib")
+else()
+    # on other platforms its just standard prefixes and suffix
+    set(PREFIX_TO_USE ${CMAKE_STATIC_LIBRARY_PREFIX})
+    set(SUFFIX_TO_USE ${CMAKE_STATIC_LIBRARY_SUFFIX})
+endif()
+
+set(EXPAT_LIBRARY ${CMAKE_CURRENT_LIST_DIR}/expat/lib/${PREFIX_TO_USE}expat${SUFFIX_TO_USE})
 set(expat_LIBRARY ${EXPAT_LIBRARY})
 
 set(EXPAT_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/expat/include)
@@ -24,7 +37,9 @@ set(EXPAT_FOUND TRUE)
 set(expat_FOUND TRUE)
 
 add_library(expat::expat STATIC IMPORTED GLOBAL)
+set_target_properties(expat::expat PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C")
 set_target_properties(expat::expat PROPERTIES IMPORTED_LOCATION ${EXPAT_LIBRARY})
+target_compile_definitions(expat::expat INTERFACE -DXML_STATIC)
 
 if (COMMAND ly_target_include_system_directories)
     # inside the O3DE ecosystem, this macro makes sure it works even in cmake < 3.19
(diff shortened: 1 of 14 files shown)


package-system/expat/test_expat_linux.sh
```diff
@@ -0,0 +1,23 @@
...
+cmake -S test -B temp/build_test -G Ninja -DCMAKE_BUILD_TYPE=Release  \
+    -DCMAKE_MODULE_PATH="$DOWNLOADED_PACKAGE_FOLDERS;$PACKAGE_ROOT" || exit 1
+
+cmake --build temp/build_test --parallel || exit 1
+
+cd test
+
+../temp/build_test/test_expat || exit 1
+
+cd ..
```
lumberyard-employee-dm, line 21: Linux supports `pushd` and `popd` as well

lumberyard-employee-dm (approved):

package-system/expat/Findexpat.cmake
```diff
@@ -0,0 +1,64 @@
...
+set(EXPAT_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/expat/include)
+set(expat_INCLUDE_DIR ${EXPAT_INCLUDE_DIR})
+
+set(EXPAT_FOUND TRUE)
+set(expat_FOUND TRUE)
+
+add_library(expat::expat STATIC IMPORTED GLOBAL)
+set_target_properties(expat::expat PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C")
+set_target_properties(expat::expat PROPERTIES IMPORTED_LOCATION ${EXPAT_LIBRARY})
+target_compile_definitions(expat::expat INTERFACE -DXML_STATIC)
```
amznestebanpapp, line 42: this define seems quite generic, the `-D` is also not needed (and is preferred to not have it in case some compiler passes defines in a different way)
If we are not using that `XML_STATIC` define, I would suggest to remove this line

package-system/expat/build_config.json
```diff
@@ -0,0 +1,143 @@
...
+                  "-G",
+                  "Xcode",
+                  "-DEXPAT_BUILD_DOCS=OFF",
+                  "-DEXPAT_BUILD_TOOLS=OFF",
+                  "-DEXPAT_BUILD_EXAMPLES=OFF",
+                  "-DEXPAT_SHARED_LIBS=OFF",
+                  "-DEXPAT_BUILD_FUZZERS=OFF",
+                  "-DEXPAT_BUILD_TESTS=OFF",
+                  "-DCMAKE_TOOLCHAIN_FILE=../../../../Scripts/cmake/Platform/Mac/Toolchain_mac.cmake",
+                  "-DBUILD_SHARED_LIBS=0"
```
amznestebanpapp, line 77: suggest giving the same ordering as the previous cases so it is easier to compare
if ordering doesnt affect anything (which should be this case), using alphabetical ordering keeps all the lists consistent

package-system/expat/test/example_file.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<rootelement>
+    <!-- a comment in here and a utf-8 element. This is not an 
+    exhaustive self test of expat though, it has those in its own code. -->
+    <special-element>✔️</special-element>
```
amznestebanpapp, line 5: this may break if we ever introduce the validation scripts into these jobs, I suggest using the escape sequence if possible

amznestebanpapp (approved):
looks awesome to me

package-system/expat/Findexpat.cmake
lawsonamzn: Ouch, yeah, its absolutely needed.  ITs what causes EXPAT's headers (ie, expat.h) to not try to dll import.

The weird thing is, this is working on every compiler and every platform we use, Jenkins can link and build them, etc... so CMake must either understand -DXML_STATIC  or -DXML_STATIC is working across every platform (?).

Should I hold out and rev the package?  Or ticket it for next time we rev expat?  Any thoughts?

package-system/expat/build_config.json
lawsonamzn: OK.  I might add some additional changes to a new PR for next time we build expat packages and so that it serves as a good example.

package-system/expat/test/example_file.xml
lawsonamzn: Escape sequence?   This is a utf-8 encoded file with utf text inside it.   Woudln't escaping it reduce the value of making sure expat can read actual utf-8 encoded files?

package-system/expat/Findexpat.cmake
amznestebanpapp: Unfortunately abotu the name of the define, just something to be careful about.

Yes, CMake, removes the `-D`: https://cmake.org/cmake/help/latest/command/target_compile_definitions.html
Fine to do it in a future change, is totally minor/cosmetic

package-system/expat/build_config.json
amznestebanpapp: sure, I am fine with another PR

package-system/expat/test/example_file.xml
amznestebanpapp: I see... I was thinking that is not a great test for our case because we dont allow these characters in xml files (we have a validation check). 
I am ok with leaving it, just something to be aware if we want to run the validation scripts in this repo (those validations may be useful since we have the legal header validations there)

package-system/expat/Findexpat.cmake
lawsonamzn: whew, saved by cmake maintainers / developers :)

spham-amzn (approved):

amznestebanpapp (approved):

PR #81, opened 2022-01-25
Add vulkan validation layer for Windows and Linux.

AMZN-tpeng:
Signed-off-by: Peng <tonypeng@amazon.com>

### DIFF
--- a/package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
+++ b/package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
@@ -1,0 +1,40 @@
+#
+# All or portions of this file Copyright (c) Amazon.com, Inc. or its affiliates or
+# its licensors.
+#
+# For complete copyright and license terms please see the LICENSE at the root of this
+# distribution (the "License"). All use of this software is governed by the License,
+# or, if provided, by the license below or the license accompanying this file. Do not
+# remove or modify any license notices. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#
+
+set(MY_NAME "vulkan-validationlayers")
+
+set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
+if (TARGET ${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+
+if (${PAL_PLATFORM_NAME} STREQUAL "Windows")
+    set(PATH_TO_DLL ${CMAKE_CURRENT_LIST_DIR}/vulkan-validationlayers/lib/release)
+    
+    set(${MY_NAME}_RELEASE_RUNTIME_DEPENDENCIES
+        ${PATH_TO_DLL}/VkLayer_khronos_validation.dll
+    )
+
+    set(${MY_NAME}_RUNTIME_JSON_DEPENDENCIES
+        ${PATH_TO_DLL}/VkLayer_khronos_validation.json
+    )
+
+    ly_add_target_files(
+        TARGETS 
+        ${TARGET_WITH_NAMESPACE} 
+        FILES 
+        ${${MY_NAME}_RELEASE_RUNTIME_DEPENDENCIES} 
+        ${${MY_NAME}_RUNTIME_JSON_DEPENDENCIES})
+endif()
+
+set(${MY_NAME}_FOUND True)
--- a/package-system/vulkan-validationlayers/build_config.json
+++ b/package-system/vulkan-validationlayers/build_config.json
@@ -1,0 +1,59 @@
+{
+    "git_url":"https://github.com/KhronosGroup/Vulkan-ValidationLayers.git",
+    "git_tag":"v1.2.198",
+    "package_name":"vulkan-validationlayers",
+    "package_version":"1.2.198",
+    "package_url":"https://vulkan.lunarg.com/",
+    "package_license":"Apache-2.0",
+    "package_license_file":"LICENSE.txt",
+    "cmake_find_source":"Findvulkan-validationlayers.cmake",
+    "cmake_find_target":"Findvulkan-validationlayers.cmake",
+    "cmake_src_subfolder": "vulkan-validation",
+    "cmake_build_args" : [ "-j", "8"],
+    "cmake_generate_args_common": [],
+    "cmake_build_args_common": [
+        "--parallel"
+    ],
+    "Platforms":{
+        "Windows":{
+            "Windows":{   
+				"custom_build_cmd" : [
+                    "build_vulkan_validation_windows.cmd"
+                ],
+                "custom_install_cmd": [
+                    "install_vulkan_validation_windows.cmd"
+                ]
+            },
+			"Android":{
+                "custom_build_cmd" : [
+                ],
+                "custom_install_cmd": [
+                ]
+            }
+        },
+		"Darwin":{
+            "Mac":{
+                "custom_build_cmd": [
+                 ],
+                 "custom_install_cmd": [
+                 ]
+            },
+            "iOS":{
+                "custom_build_cmd": [
+                 ],
+                 "custom_install_cmd": [
+                 ]
+             }
+        },
+        "Linux":{
+            "Linux":{
+                "custom_build_cmd": [
+                ],
+                "custom_install_cmd": [
+                ],
+                "custom_test_cmd" : [
+               ]
+            }
+        }
+    }
+}
--- a/package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
+++ b/package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
@@ -1,0 +1,49 @@
+@rem #
+@rem # Copyright (c) Contributors to the Open 3D Engine Project.
+@rem # For complete copyright and license terms please see the LICENSE at the root of this distribution.
+@rem # 
+@rem # SPDX-License-Identifier: Apache-2.0 OR MIT
+@rem #
+
+@rem # note that we explicitly turn off the compilation of all features that rely on 3rd Party Libraries
+@rem # except the ones we want.  This prevents the cmake build system from automatically finding things
+@rem # if they happen to be installed locally, which we don't want.
+
+@rem # cmake expects fowardslashes:
+set "DOWNLOADED_PACKAGE_FOLDERS=%DOWNLOADED_PACKAGE_FOLDERS:\=/%"
+
+@rem cd temp/build
+
+@rem release build
+C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config release
+cmake -C temp/src/external/helper.cmake -S temp/src -B temp/build
+cmake --build temp/build --config Release --target clean
+cmake --build temp/build --config Release --target install 
+mkdir temp\\build\\install\\lib\\release
+move temp\\build\\install\\lib\\* temp\\build\\install\\lib\\release\\
+
+@rem debug build
+C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config debug
+cmake --build temp/build --config Debug --target clean
+cmake -C temp/src/external/helper.cmake -S temp/src -B temp/build
+cmake --build temp/build --config Debug --target install 
+mkdir temp\\build\\install\\lib\\debug
+move temp\\build\\install\\lib\\* temp\\build\\install\\lib\\debug\\
+
+@rem cmake --build . --config Release --target clean
+@rem cmake -C ../src/external/helper.cmake -DCMAKE_BUILD_TYPE=Release ../src
+@rem cmake --build . --config Release --target install
+@rem mkdir install\\lib\\release
+@rem move install\\lib\\* install\\lib\\release\\
+
+@rem cd ..\\..
+@rem C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config debug
+@rem cd temp/build
+@rem 
+@rem cmake --build . --config Debug --target clean
+@rem cmake -C ../src/external/helper.cmake -DCMAKE_BUILD_TYPE=Debug ../src
+@rem cmake --build . --config Debug --target install
+@rem mkdir install\\lib\\debug
+@rem move install\\lib\\* install\\lib\\debug\\
+
+exit /b 0
--- a/package-system/vulkan-validationlayers/install_vulkan_validation_windows.cmd
+++ b/package-system/vulkan-validationlayers/install_vulkan_validation_windows.cmd
@@ -1,0 +1,24 @@
+@rem #
+@rem # Copyright (c) Contributors to the Open 3D Engine Project.
+@rem # For complete copyright and license terms please see the LICENSE at the root of this distribution.
+@rem # 
+@rem # SPDX-License-Identifier: Apache-2.0 OR MIT
+@rem #
+@rem #
+@setlocal
+
+set OUT_PATH=%TARGET_INSTALL_ROOT%
+set SRC_PATH=temp\\src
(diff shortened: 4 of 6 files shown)


package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
```diff
@@ -0,0 +1,49 @@
...
+@rem # except the ones we want.  This prevents the cmake build system from automatically finding things
+@rem # if they happen to be installed locally, which we don't want.
+
+@rem # cmake expects fowardslashes:
+set "DOWNLOADED_PACKAGE_FOLDERS=%DOWNLOADED_PACKAGE_FOLDERS:\=/%"
+
+@rem cd temp/build
+
+@rem release build
+C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config release
```
jeremyong-az: Hardcoded python paths

package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
```diff
@@ -0,0 +1,49 @@
...
+cmake -C temp/src/external/helper.cmake -S temp/src -B temp/build
+cmake --build temp/build --config Release --target clean
+cmake --build temp/build --config Release --target install 
+mkdir temp\\build\\install\\lib\\release
+move temp\\build\\install\\lib\\* temp\\build\\install\\lib\\release\\
+
+@rem debug build
+C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config debug
```
jeremyong-az: Hardcoded python paths

package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
```diff
@@ -0,0 +1,49 @@
...
+move temp\\build\\install\\lib\\* temp\\build\\install\\lib\\debug\\
+
+@rem cmake --build . --config Release --target clean
+@rem cmake -C ../src/external/helper.cmake -DCMAKE_BUILD_TYPE=Release ../src
+@rem cmake --build . --config Release --target install
+@rem mkdir install\\lib\\release
+@rem move install\\lib\\* install\\lib\\release\\
+
+@rem cd ..\\..
+@rem C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config debug
```
jeremyong-az: Hardcoded python paths (did you want to remove these comments instead?)

package-system/vulkan-validationlayers/build_config.json
```diff
@@ -0,0 +1,59 @@
...
+    "cmake_src_subfolder": "vulkan-validation",
+    "cmake_build_args" : [ "-j", "8"],
+    "cmake_generate_args_common": [],
+    "cmake_build_args_common": [
+        "--parallel"
+    ],
+    "Platforms":{
+        "Windows":{
+            "Windows":{   
+				"custom_build_cmd" : [
```
jeremyong-az, line 20: Nit: spacing

package-system/vulkan-validationlayers/build_config.json
```diff
@@ -0,0 +1,59 @@
...
+                    "build_vulkan_validation_windows.cmd"
+                ],
+                "custom_install_cmd": [
+                    "install_vulkan_validation_windows.cmd"
+                ]
+            },
+			"Android":{
```
jeremyong-az, line 27: Nit: spacing (and other lines in this file)

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
```diff
@@ -0,0 +1,40 @@
+#
+# All or portions of this file Copyright (c) Amazon.com, Inc. or its affiliates or
```
jeremyong-az, line 2: Copyright should be attributed to O3DE

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
```diff
@@ -0,0 +1,40 @@
...
+set(MY_NAME "vulkan-validationlayers")
+
+set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
+if (TARGET ${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+
+if (${PAL_PLATFORM_NAME} STREQUAL "Windows")
```
spham-amzn, line 21: I don't see anything refering the the Linux libraries

package-system/vulkan-validationlayers/build_config.json
```diff
@@ -0,0 +1,59 @@
...
+        },
+        "Linux":{
+            "Linux":{
+                "custom_build_cmd": [
+                ],
+                "custom_install_cmd": [
+                ],
+                "custom_test_cmd" : [
+               ]
+            }
```
spham-amzn, line 56: There are no scripts to build Linux yet, so this PR is just for Windows right?

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
```diff
@@ -0,0 +1,40 @@
...
+#
+
+set(MY_NAME "vulkan-validationlayers")
+
+set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
+if (TARGET ${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
```
lawsonamzn, line 19: It doesn't seem like this is an INTERFACE library.  It looks like a library which has a DLL, which would be a 
SHARED IMPORTED GLOBAL
not
INTERFACE IMPORTED GLOBAL

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake
```diff
@@ -0,0 +1,40 @@
...
+        ${PATH_TO_DLL}/VkLayer_khronos_validation.json
+    )
+
+    ly_add_target_files(
+        TARGETS 
+        ${TARGET_WITH_NAMESPACE} 
+        FILES 
+        ${${MY_NAME}_RELEASE_RUNTIME_DEPENDENCIES} 
+        ${${MY_NAME}_RUNTIME_JSON_DEPENDENCIES})
+endif()
```
lawsonamzn, line 38: This would add the dependencies for copies, I think, but in CMAKE terms the corect way to do it is to set
IMPORTED_LOCATION property on the SHARED IMPORTED GLOBAL library.
in otehr words
```
set(_DLL_NAME ${PATH_TO_DLL}/VkLayer_khronos_validation${CMAKE_SHARED_LIBRARY_SUFFIX})
# ${CMAKE_SHARED_LIBRARY_SUFFIX} is '.dll' on windows but appropriately named on other systems.

# Declare that its a SHARED library (ie, a dll) as opposed to a STATIC (lib) or INTERFACE (header only)
# declaring it as IMPORTED means we are going to ingest a pre-existing thing, rather than build one.
add_library(${TARGET_WITH_NAMESPACE} SHARED IMPORTED GLOBAL)

# declare that the primary imported location (ie, the dll for a shared, the lib for a static) is the dll
# the imported location is basically "What to import?".  For a shared library, its the dll.
set_target_properties(${TARGET_WITH_NAMESPACE} PROPERTIES IMPORTED_LOCATION ${_DLL_NAME})

# files that cmake doesn't care about / are not involved in the build system have to be handled custom.
ly_add_target_files(
        TARGETS 
        ${TARGET_WITH_NAMESPACE} 
        FILES 
         ${${MY_NAME}_RUNTIME_JSON_DEPENDENCIES})
```

package-system/vulkan-validationlayers/build_config.json
```diff
@@ -0,0 +1,59 @@
+{
+    "git_url":"https://github.com/KhronosGroup/Vulkan-ValidationLayers.git",
+    "git_tag":"v1.2.198",
+    "package_name":"vulkan-validationlayers",
+    "package_version":"1.2.198",
+    "package_url":"https://vulkan.lunarg.com/",
```
lawsonamzn, line 6: Consider bieng more specific
https://github.com/KhronosGroup/Vulkan-ValidationLayers

package-system/vulkan-validationlayers/build_config.json
```diff
@@ -0,0 +1,59 @@
...
+    "package_license":"Apache-2.0",
+    "package_license_file":"LICENSE.txt",
+    "cmake_find_source":"Findvulkan-validationlayers.cmake",
+    "cmake_find_target":"Findvulkan-validationlayers.cmake",
+    "cmake_src_subfolder": "vulkan-validation",
+    "cmake_build_args" : [ "-j", "8"],
+    "cmake_generate_args_common": [],
+    "cmake_build_args_common": [
```
lawsonamzn, line 14: These fields are unused if a custom build command is used (in which case INSTEAD of calling cmake, it just invokes the cmd file instead)

lawsonamzn (commented):
Of course, this would need the other platforms before its complete (or O3DE would have to have no package association for those platforms if you wan to to do it iteratively)

package-system/vulkan-validationlayers/build_config.json
AMZN-tpeng: Linux and Android are just stubs now since they use vulkan on Atom as well. ly_add_target() in the Atom.RHI.Vulkan cmake will have 3rdParty::vulkan-validation-layers.

package-system/vulkan-validationlayers/build_config.json
AMZN-tpeng: If I don't have these entries, errors such as:
Required key 'cmake_generate_args_common' not found in build config

will show up. What would be the correct way of handling this?

moudgils (approved):

package-system/vulkan-validationlayers/build_config.json
spham-amzn: That shouldn't be the case. For example, the build_config.json for sqlite does not have this key at all because it only uses custom build cmds

package-system/vulkan-validationlayers/build_config.json
spham-amzn: If you intend on creating 'stub' packages, then i'd suggest not sharing the same 'Findvulkan-validationlayers.cmake' as the source file. Instead create a 'Findvulkan-validationlayers.cmake.stub' as the 'cmake_find_source' for those platforms, and have that find just declare the empty library, and add comments there saying that it is not supported yet. 

Or even simpler, use PAL to only add a dependency to atom on windows for this package, and don't declare anything for platforms other than windows. So you would only need to push 1 package instead of stubs.

package-system/vulkan-validationlayers/build_config.json
AMZN-tpeng: I tried running the script on SQLite and got similar error. Maybe I am running the wrong script or wrong arguments?

PS E:\3p\3p-package-source> C:\Users\tonypeng\AppData\Local\Programs\Python\Python38\python.exe ..\3p-package-scripts\o3de_package_scripts\build_package.py --search_path E:\3p\3p-package-source SQLite-3.37.2-rev1-windows
Output folder for packages is 'E:\3p\3p-package-source\packages' - Override with --output_folder
Loading windows package lists from E:\3p\3p-package-source...
Loading and merging E:\3p\3p-package-source\package_build_list_host_windows.json ...
Calling build script: "E:\3p\3p-package-source\Scripts\extras\pull_and_build_from_git.py ../../package-system/sqlite --platform-name Windows --package-root ../../package-system --clean"...
Command: cmake --version
Detected CMake: cmake version 3.20.0
Required key 'cmake_src_subfolder' not found in build config
Package SQLite-3.37.2-rev1-windows failed to build from source.

lawsonamzn: Sounds like a bug, maybe I introduced one.  Its not supposed to be a required key.  So its a bug in the script to require it.

package-system/vulkan-validationlayers/build_config.json
AMZN-tpeng: Added stubs for linux and android.

package-system/vulkan-validationlayers/build_config.json
AMZN-tpeng: I think we should fix pull_and_build_from_git.py in a separate PR to have a better track of its history.

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake.stub
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+# Blank stub for Linux and Android as they are not supported yet.
```
spham-amzn, line 9: You will still need to declare the library I think.
```suggestion

set(MY_NAME "vulkan-validationlayers")

set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
if (TARGET ${TARGET_WITH_NAMESPACE})
    return()
endif()

# Blank stub library for Linux and Android as they are not supported yet.
add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)

set(${MY_NAME}_FOUND True)
```

package-system/vulkan-validationlayers/build_config.json
spham-amzn: PR for that fix is here: https://github.com/o3de/3p-package-source/pull/82

AMZN-tpeng: Took out the fixed optional arguments.

package-system/vulkan-validationlayers/Findvulkan-validationlayers.cmake.stub
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+set(MY_NAME "vulkan-validationlayers")
+set(TARGET_WITH_NAMESPACE "3rdParty::${MY_NAME}")
+if (TARGET ${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+# Blank stub library for Linux and Android as they are not supported yet.
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
```
lawsonamzn, line 14: The stub seems like a bad way to do it, irght? it either means having to rev the windows package later when you support these, or, having to ship empty stub packages to people unnecessraily.

Instead of doing a stub, you could put the dependency and declaration of ly_associate-package on 3rdParty::vulkan-validationlayers in O3DE source in a PAL file where its needed and only on the platforms its needed on.  You could even define something in there if you needed code to know whether its there or not.

package-system/vulkan-validationlayers/build_vulkan_validation_windows.cmd
```diff
@@ -0,0 +1,33 @@
...
+@rem # except the ones we want.  This prevents the cmake build system from automatically finding things
+@rem # if they happen to be installed locally, which we don't want.
+
+@rem # cmake expects fowardslashes:
+set "DOWNLOADED_PACKAGE_FOLDERS=%DOWNLOADED_PACKAGE_FOLDERS:\=/%"
+
+@rem cd temp/build
+
+@rem release build
+python.exe temp/src/scripts/update_deps.py --dir temp/src/external --arch x64 --config release
```
lawsonamzn, line 18: If you want to use the same python that pull_and_build_from_git was launched you can use the env var %PYTHON_BINARY%
pull_and_build_from_git.py sets this env var to the python binary that was used to invoke it in all child processes.  This ensures that even if python.exe is not on the path, or its the wrong version, you will be invoked with the approrpiate python3 executable.  It does need python3, right?  Its not a python2 script?

lawsonamzn:
If you need help with this (especially on the o3de cmake side) to do this without stubs, let me know.   Its possible to write scripts on the O3DE side that check if a package is associated for this platform and set defines and things if it is or isn't, as well as declare package associations conditionally (on platform or other)

PR #88, opened 2022-02-04
Update lua to 5.4.4 (fixes #7267)

lawsonamzn:
New version of LUA (package building scripts).  Note that the new package is uplaoded on dev, and a PR to the o3de repo will be forthcoming.


lumberyard-employee-dm (approved):

lsemp3d (approved):

spham-amzn (approved):

lawsonamzn:
So this worked in all the current platforms in local testing, but went red on the build farm.  On inspection though, the problem appears to be an unrelated Atom shader compiler issue.  While we resolve this, I'll work to stage this to update once it goes green.

PR #89, opened 2022-02-07
Upgrade DXC to v1.6.2112Upgrade DXC to v1.6.2112

galibzon:
The o3de branch for dxc 1.6.2112, which contains all the o3de customizations is located here:
https://github.com/o3de/DirectXShaderCompiler/tree/release-1.6.2112-o3de

Signed-off-by: galibzon <66021303+galibzon@users.noreply.github.com>


moudgils (approved):

package-system/DirectXShaderCompiler/build_dxc_windows.cmd
```diff
@@ -22,7 +22,7 @@ IF %ERRORLEVEL% NEQ 0 (
 )
 
 REM Make sure TAEF is installed
-call utils\hct\hctgettaef.py
```
lawsonamzn, line 25: I think the %PYTHON_BINARY% define is declared by pull_and_build_from_git in all build scripts if you want this to work even without python in the path.   Its better than assuming that python is on the path, or that its the correct one (ie, might be python 27)

galibzon: Thanks for the feedback. I changed it to %PYTHON_BINARY%.

package-system/DirectXShaderCompiler/build_config.json
```diff
@@ -50,8 +52,10 @@
                "-DSPIRV_BUILD_TESTS:BOOL=ON"
             ],
             "cmake_build_args":[
+               "--target",
+               "all llvm-as llvm-dis",
                "-j",
-               "8"
+               "16"
```
lawsonamzn, line 58: I'd recommend specifying 
```
--parallel
```

instead of using the -j parameter.  (and don't include a number)


instead.  This will pass the necessary command line parameters down to the builder and will work completely independently of OS or builder.  For example for gcc it will send -j(cores) and for ninja it will omit it since its the default, and so on

lawsonamzn (approved):
You can ship it as is, or ship it with --parallel, it won't change the package output, so packages dont need to be reuploaded or anything.

package-system/DirectXShaderCompiler/build_config.json
lawsonamzn: so an example would be to replace
```
"-j",
"16"
```
with just
```
"--parallel"
```

galibzon: Thanks for the tip. changed to --parallel

PR #90, opened 2022-02-09
Update SQLite3 for Mac to fix version mismatch

spham-amzn:
The current uploaded version of SQLite3 for Mac is actually 3.32, not 3.37. The binary was rebuilt and verified, but a rev bump is needed (just for Mac)

Signed-off-by: spham <82231385+spham-amzn@users.noreply.github.com>


lawsonamzn:
looks good but how come this fixes it, if no changes made to the patch build?  What caused it in the first place?

spham-amzn:
> Why rev a package when there's no changes to the package build script besides rev-ing the package? Won't the same problem manifest again?

So the root cause (which should be fixed by the update in pull_and_build_from_git.py) was that there was a pre-existing SQLite-mac working package folder from the previous version that I had built (3.32). When I built 3.37 for Mike, the scripts to assemble the package did a `cp` command that did not overwrite the existing files (I did not catch any message) so the pre-existing SQLite-mac package file was tar'd up instead. 

The intermediate package folder should be clean anyways to prevent any unwanted artifacts to appear.

lawsonamzn (approved):

galibzon (approved):

amzn-mike (approved):

PR #93, opened 2022-02-15
Add gamelift server sdk build script

onecent1101:
## Details
Added windows and linux platform build script for gamelift server sdk

## Testing
Tested with both windows and linux server launcher locally

Signed-off-by: onecent1101 <liug@amazon.com>


package-system/AWSGameLiftServerSDK/README.md
```diff
@@ -3,115 +3,30 @@
...
-* Either of the following:  
-  * Microsoft Visual Studio 2017 - Windows Build
-  * Clang 6.0 - Linux Build
-* CMake version 3.1 or later
-* Python version 3.6 or later
+* Either of the following:
+  * Microsoft Visual Studio 2019 - Windows Build
+  * Clang 12.0 and GCC 9.3 - Linux Build
+* CMake version 3.20 or later
+* Python version 3.7 or later
```
lmbr-pip, line 10: 3.7.12 I believe is the minimum version of python we support?

package-system/AWSGameLiftServerSDK/README.md
```diff
@@ -3,115 +3,30 @@
...
 ## Minimum requirements:
-* Either of the following:  
-  * Microsoft Visual Studio 2017 - Windows Build
-  * Clang 6.0 - Linux Build
-* CMake version 3.1 or later
-* Python version 3.6 or later
+* Either of the following:
+  * Microsoft Visual Studio 2019 - Windows Build
+  * Clang 12.0 and GCC 9.3 - Linux Build
+* CMake version 3.20 or later
```
lmbr-pip, line 9: CMake 3.20.5 or later to be precise.

Assume some of this is driven off Linux min specs for O3DE. Should we link to those docs?

package-system/AWSGameLiftServerSDK/README.md
```diff
@@ -3,115 +3,30 @@
...
+$ sudo ln -s /usr/lib/llvm-12/bin/clang++ /usr/bin/clang++ 
+```
 
 ## FAQ
-* For windows build, use `--msbuild_path` argument to specify msbuild executable path if msbuild is unrecognized as a valid command, like
-  ```
-  & python build_and_package.py --config gamelift-sdk.json --platform windows --version 3.4.1 --msbuild_path "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\MSBuild.exe"
-  ```
-
+* For windows build, as dependent boost version is old and not well compatible with Visual Studio 2019, feel free to switch to `Visual Studio 15 2017` if necessary
```
lmbr-pip, line 28: Avoid ambiguous instructions. Can then build with VS2019 or is the recommendation to use VS2017? Maybe "Because the Amazon GameLift Server SDK depends on an older version of boost, it is recommend that you build to VS2017" or similar. Unclear what problems they may encounter or when then should switch.

package-system/AWSGameLiftServerSDK/README.md
```diff
@@ -3,115 +3,30 @@
...
-* For windows build, use `--msbuild_path` argument to specify msbuild executable path if msbuild is unrecognized as a valid command, like
-  ```
-  & python build_and_package.py --config gamelift-sdk.json --platform windows --version 3.4.1 --msbuild_path "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\MSBuild.exe"
-  ```
-
+* For windows build, as dependent boost version is old and not well compatible with Visual Studio 2019, feel free to switch to `Visual Studio 15 2017` if necessary 
+  (Refer to function `configure_sdk_project` in `build_package_image.py` script)
 * For windows build, if you get error message `The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters`
-  Please move `GameLift-Cpp-ServerSDK-{version-number}` directory out of nested directories and retry.
+  Please move `3p-package-source` directory out of nested directories and retry.
```
lmbr-pip, line 31: As this is a well known issue, are there Msoft docs we can link to here?

nit: move out of nested directories to me seems hard to parse. Please move  directory to shorter path location and retry?

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+
+PACKAGE_NAME: str = "AWSGameLiftServerSDK"
+PACKAGE_VERSION: str = "3.4.2-rev1"
+PACKAGE_PLATFORM: str = ""
+PACKAGE_URL: str = "https://aws.amazon.com/documentation/gamelift/"
+PACKAGE_LICENSE: str = "Apache-2.0"
+PACKAGE_LICENSE_FILE: str = "LICENSE_AMAZON_GAMELIFT_SDK.TXT"
+
+GAMELIFT_SERVER_SDK_RELEASE_VERSION: str = "4.0.2"
+GAMELIFT_SERVER_SDK_DOWNLOAD_URL: str = "https://gamelift-release.s3-us-west-2.amazonaws.com/GameLift_06_03_2021.zip"
```
lmbr-pip, line 29: Can't believe this is still latest, but it is.

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+def subp_args(args) -> str:
+    arg_string = " ".join([arg for arg in args])
+    print(f"Command: {arg_string}")
+    return arg_string
+
+def delete_folder(folder: pathlib.Path) -> None:
+    shutil.rmtree(folder.resolve())
+
+# steps
+def collect_package_info() -> None:
```
lmbr-pip, line 59: Default to providing comments on all python functions

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+
+        target_cmake_file: pathlib.Path = working_directory.output_path.joinpath(f"Find{PACKAGE_NAME}.cmake")
+        target_cmake_file.write_text(find_cmake_content)
+    else:
+        raise Exception(f"Error finding cmake source file: {cmake_file_source.resolve()}")
+
+
+if __name__ == '__main__':
+    try:
+        print("Collecting package build info for GameLift Server SDK...")
```
lmbr-pip, line 253: Prefer python logging over print

lmbr-pip (commented):
LG.

Really appreciate your hardwork here. Python code is very clean.

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+                        working_directory.libs_output_path.joinpath("include").resolve())
+        shutil.copy2(working_directory.source_path.joinpath(PACKAGE_LICENSE_FILE).resolve(),
+                     working_directory.libs_output_path.resolve())
+        shutil.copy2(working_directory.source_path.joinpath("NOTICE_C++_AMAZON_GAMELIFT_SDK.TXT").resolve(),
+                     working_directory.libs_output_path.resolve())
+        
+        print("Generating package info into output directory...")
+        generate_packageInfo(working_directory)
+        
+        print("Gnerating cmake file into output directory...")
```
lmbr-pip, line 284: "Gnerating" -> Generating

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+        delete_folder(build_output_directory)
+    
+    build_libs_output_directory: pathlib.Path = build_output_directory.joinpath(f"{PACKAGE_NAME}")
+    build_libs_output_directory.mkdir(parents=True)
+    
+    return WorkingDirectoryInfo(root_directory, source_directory, build_directory, build_output_directory, build_libs_output_directory)
+
+def download_gamelift_server_sdk(working_directory: WorkingDirectoryInfo) -> None:
+    # download sdk from url
+    gamelift_sdk_zip_file: str = working_directory.root_path.joinpath("gamelift_server_sdk.zip").resolve()
```
lmbr-pip, line 95: Nit: resolve returns a path not a string. Be careful when you are using implicit conversion like this, probably should be wrapped in str()

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+        choices=PACKAGE_PLATFORM_OPTIONS
+    )
+    
+    args = parser.parse_args()
+    global PACKAGE_PLATFORM
+    PACKAGE_PLATFORM = args.platform_name
+
+def prepare_working_directory() -> WorkingDirectoryInfo:
+    # root directory
+    root_directory: pathlib.Path = PACKAGE_BASE_PATH.joinpath("temp")
```
junbo75, line 73: Nit: May worth making the temp path a CLI argument people can specify in the configuration file instead of hardcoding it here.

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+    
+    build_libs_output_directory: pathlib.Path = build_output_directory.joinpath(f"{PACKAGE_NAME}")
+    build_libs_output_directory.mkdir(parents=True)
+    
+    return WorkingDirectoryInfo(root_directory, source_directory, build_directory, build_output_directory, build_libs_output_directory)
+
+def download_gamelift_server_sdk(working_directory: WorkingDirectoryInfo) -> None:
+    # download sdk from url
+    gamelift_sdk_zip_file: str = working_directory.root_path.joinpath("gamelift_server_sdk.zip").resolve()
+    with urllib.request.urlopen(GAMELIFT_SERVER_SDK_DOWNLOAD_URL) as response, open(gamelift_sdk_zip_file, 'wb') as out_file:
```
junbo75, line 96: Do we need to handle the exception that the URL is invalid here?

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+                                cwd=source_folder)
+    if git_result.returncode != 0:
+        raise Exception(f"Error git init sdk source: {str(git_result.stderr)}")
+    git_result = subprocess.run(subp_args(["git","add","."]), 
+                                shell=True,
+                                capture_output=True,
+                                cwd=source_folder)
+    if git_result.returncode != 0:
+        raise Exception(f"Error git add sdk source: {str(git_result.stderr)}")
+    git_result = subprocess.run(subp_args(["git","commit","-m","temp"]),
```
junbo75, line 118: Curious: What are the changes we're committing here?

package_build_list_host_linux.json
```diff
@@ -5,6 +5,7 @@
     "comment4" : "Note:  Build from source occurs before build_from_folder",
     "build_from_source": {
         "assimp-5.1.6-rev1-linux":  "Scripts/extras/pull_and_build_from_git.py ../../package-system/assimp --platform-name Linux --package-root ../../package-system --clean",
+        "AWSGameLiftServerSDK-3.4.2-rev1-linux": "package-system/AWSGameLiftServerSDK/build_package_image.py --platform-name linux",
```
junbo75, line 8: Same question as above: Seems like GAMELIFT_SERVER_SDK_RELEASE_VERSION is 4.0.2. Then why are we specifying 3.4.2 here?

package_build_list_host_windows.json
```diff
@@ -5,6 +5,7 @@
     "comment4" : "Note:  Build from source occurs before build_from_folder",
     "build_from_source": {
         "assimp-5.1.6-rev1-windows":  "Scripts/extras/pull_and_build_from_git.py ../../package-system/assimp --platform-name Windows --package-root ../../package-system --clean",
+        "AWSGameLiftServerSDK-3.4.2-rev1-windows": "package-system/AWSGameLiftServerSDK/build_package_image.py --platform-name windows",
```
junbo75, line 8: Same question as above: Seems like GAMELIFT_SERVER_SDK_RELEASE_VERSION is 4.0.2. Then why are we specifying 3.4.2 here?

package-system/AWSGameLiftServerSDK/AWSGameLiftServerSDK-3.4.2-windows.patch
```diff
@@ -0,0 +1,27 @@
...
+index 3333da3..2ef3126 100644
+--- a/cmake/External_sioclient.cmake
++++ b/cmake/External_sioclient.cmake
+@@ -54,7 +54,8 @@ elseif(WIN32)
+      if(MSVC14)
+         add_custom_command(
+             TARGET sioclient-src POST_BUILD
+-            COMMAND git pull origin master
++            COMMAND git fetch origin
++            COMMAND git merge 56123c87598f8b1dd471be83ca841ceae07f95ba
```
junbo75, line 24: May worth adding a comment about the reason why we need to merge this commit.

package-system/AWSGameLiftServerSDK/README.md
```diff
@@ -3,115 +3,30 @@
...
-   ```
-6. Update `package-system/AWSGameLiftServerSDK/windows/PackageInfo.json` with built version `PackageName`
-
-7. Update `<3p-package-source path>/package_build_list_host_windows.json` with updated `PackageName` in step 6
+## Prepare steps:
+1. As there is no public repository for server sdk source code, please double check [Amazon GameLift Release Notes](https://docs.aws.amazon.com/gamelift/latest/developerguide/release-notes.html)
+   to confirm server sdk version information
+2. Modify `PACKAGE_VERSION`, `GAMELIFT_SERVER_SDK_RELEASE_VERSION` and `GAMELIFT_SERVER_SDK_DOWNLOAD_URL` to update version
 
 ## Build the GameLiftServerSdk - Linux
```
junbo75, line 18: This section only talks about how to link Clang compiler for cmake build but doesn't provide instructions to actually "build" the SDK. May worth rename the title here.

onecent1101: will update

package-system/AWSGameLiftServerSDK/README.md
onecent1101: will update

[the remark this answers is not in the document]
onecent1101: will update

package-system/AWSGameLiftServerSDK/README.md
onecent1101: will update

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -0,0 +1,289 @@
...
+import os
+import pathlib
+import subprocess
+import shutil
+import urllib.request
+import zipfile
+from typing import List
+
+PACKAGE_NAME: str = "AWSGameLiftServerSDK"
+PACKAGE_VERSION: str = "3.4.2-rev1"
```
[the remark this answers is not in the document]
onecent1101: 4.0.2 is gamelift server sdk version
3.4.2 is gamelift server cpp sdk version
and this way is following existing format which requires to update version inside build script, refer to https://github.com/o3de/3p-package-source/blob/main/package_build_list_host_windows.json

we can use argument, but for our use case that requires too many argument to pass, like gamelift server sdk version, cpp sdk version, release date and platform

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: this is the default behavior of build from git script

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: will update

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: build script will catch the exception

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: will update

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: it is for apply patch file if the directory is not a git workspace initially

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: will update

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: will update

package-system/AWSGameLiftServerSDK/README.md
allisaurus: Do you plan to add instructions for how to "build" here later? Or is the idea that this will now happen automatically when building the server target?

onecent1101: There is no specific build instruction, as it follows the general workflow of 3p system

package-system/AWSGameLiftServerSDK/build_package_image.py
onecent1101: going to keep print, tried logging, but it just overcomplicated the use case here.
The purpose is just print out to terminal, and logging requires extra step to setup handler for doing it.

lawsonamzn (approved):
This looks fine to me, TBH - but surprised that we bother making non-static versions.  Could we not just make static only versions and use them whether we're in monolithic or not?

onecent1101:
> This looks fine to me, TBH - but surprised that we bother making non-static versions. Could we not just make static only versions and use them whether we're in monolithic or not?

yep, we have both shared and static version right now with this change.

junbo75 (approved):
LG

PR #94, opened 2022-02-16
Upgrade ilmbase to OpenEXR.

hershey5045:
Notes:
* Upgrade ilmbase to OpenEXR 3.1.3
* Remove old custom ilmbase/OpenEXR cmake files and instead rely on the cmake files within the OpenEXR package. 
* Add extra_files_to_copy (which takes a list of pairs<source, destination>) in pull_and_build_from_git.py
* Tested on windows/mac/linux
* Green on origin jenkins

Signed-off-by: hershey5045 <43485729+hershey5045@users.noreply.github.com>


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -779,6 +779,19 @@ def check_build_keys(self, keys_to_check):
...
+        """
+        Copies any extra files specified in the build config into the destination folder for packaging.
+        """
+        extra_files_to_copy = self.platform_config.get('extra_files_to_copy', [])
+
+        for (source, dest) in extra_files_to_copy:
+            print(f"Source file: {self.base_folder / source}, Destination file: {self.package_install_root / dest}")
+            shutil.copy2(
+                self.base_folder / source,
+                self.package_install_root / dest
```
lawsonamzn, line 792: This  needs error checking and auto-failure if the copy2 fails.  Or does it throw an uncaught exception (which would be okay)?  Maybe try it with a file that doesn't exist and see what happens, we want the package to 'fail'

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -919,6 +932,9 @@ def execute(self):
             # Build the package
             self.build_for_platform()
 
+            # Copy extra files specified in the build config
+            self.copy_extra_files()
+
```
lawsonamzn, line 937: Should this really be in the "if" statement?  Feels like it should happen regardless of how it gets the build.

package-system/OpenEXR/build_config.json
```diff
@@ -1,100 +1,83 @@
...
+               "-DPYTHON=OFF",
+               "-DCMAKE_INSTALL_PREFIX=\"../../OpenEXR-mac/OpenEXR\"",
+               "-DCMAKE_BUILD_TYPE=Release"
+            ],
+            "custom_test_cmd" : [
+               "./test_openexr_mac.sh"
+			],
+			"extra_files_to_copy": [
+				["FindImath.cmake", "FindImath.cmake"]
+			]
```
spham-amzn, line 59: nit: text alignment seems off

package-system/OpenEXR/build_config.json
```diff
@@ -1,100 +1,83 @@
...
+               "-DBUILD_TESTING=OFF",
+               "-DOPENEXR_RUN_FUZZ_TESTS=OFF",
+               "-DOPENEXR_INSTALL_EXAMPLES=OFF",
+               "-DOPENEXR_BUILD_TOOLS=OFF",
+               "-DBUILD_SHARED_LIBS=OFF",
+               "-DCMAKE_INSTALL_PREFIX=\"../../openexr-linux/OpenEXR\""
+            ],
+			"extra_files_to_copy": [
+				["FindImath.cmake", "FindImath.cmake"]
+			]
```
spham-amzn, line 80: nit: test alignment seems off

Scripts/extras/pull_and_build_from_git.py
spham-amzn, line 937: Also, this functionality should go inside the `self.generate_package_info()` function instead of here.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -779,6 +779,19 @@ def check_build_keys(self, keys_to_check):
 
         return False
 
+    def copy_extra_files(self):
+        """
+        Copies any extra files specified in the build config into the destination folder for packaging.
+        """
+        extra_files_to_copy = self.platform_config.get('extra_files_to_copy', [])
```
spham-amzn, line 786: if 'extra_files_to_copy' is a new key, you will need to document it above with the other keys

Scripts/extras/pull_and_build_from_git.py
hershey5045: The 2nd one--it will throw a FileNotFoundError which will cause the package to fail.

Scripts/extras/pull_and_build_from_git.py
hershey5045: We need the files to be copied and available before self.test_package()

hershey5045: Done!

package-system/OpenEXR/build_config.json
hershey5045: Moved the extra_files_to_copy to global scope for this file.

Scripts/extras/pull_and_build_from_git.py
hershey5045: extra_files_to_copy moved to PackageInfo. Comment added as well for the new "extra_files_to_copy" build_config key.

spham-amzn (approved):

PR #98, opened 2022-03-03
Fix  package_build_list_host_windows.json 

spham-amzn:
Somehow this file had a bad merge (from https://github.com/o3de/3p-package-source/commit/b813732a429cb7fb6b4ad04fa70704954b2a5c4b ) that wiped out the right versions of packages to be built, so it didnt reflect actual packages that are live. This doesnt affect any live system, since the packages are published anyways, but it needs to be updated to any additions or rev-bumps done to any 3P library will be updated properly.

* Updated windows package list for 3P to reflect current versions
* Ordered Alphabetically


lumberyard-employee-dm (approved):

lawsonamzn (approved):
this was probably caused by using development as your original branch, where in this repo, main is the only one that should exist.  Can we delete development?

PR #99, opened 2022-03-03
Freetype -> 2.11.1, fix errors in pull_and_build_from_git.py

lawsonamzn:
As part of upgrading freetype to 2.11.1 some errors and omissions were
found in pull_and_build_from_git.py

Changes to Freetype:
* upgraded to latest stable
* Converted the complex build script to a patchless json file.
* Test scripts added which test the built freetype2 library in terms
  of linking/compiling against a trivial test app in the same way
  as O3DE might link and use it.
* Added official zlib support to it, using the standard o3de zlib.
* Standardized the FindFreetype.cmake - uppercase F, now it can be
  used by other projects that are not O3DE such as other 3p's,
  transparently.

Changes to pull_and_build_from_git to support the above:
* The option to do a 'custom install' is meaningless and would
  likely malfunction on windows, since the target's name is install
  on some build systems and INSTALL on others, and it IS case sensitive.
  This option has been removed and now the script always does a
  platform independent configure, build, install that will work with
  both the old and new scripts.
* The option to filter install was actually broken and did nothing
  due to the 'temp install' directory and final install directory
  getting mixed up in code.
* The install filter can now be globally specified and then overridden
  per platform, leading to simpler scripts.


rgba16f (approved):

AMZN-alexpete (approved):

cgalvan (approved):
Nice!

spham-amzn (approved):

lumberyard-employee-dm (approved):

PR #102, opened 2022-03-10
Upgrade OpenSSL to 1.1.1m

junbo75:
Signed-off-by: Junbo Liang <68558268+junbo75@users.noreply.github.com>

- upgraded to 1.1.1m
- Test scripts added which test the built OpenSSL library in terms
of linking/compiling against a trivial test app in the same way
as O3DE might link and use it.
- Standardized the FindOpenSSL.cmake - Add targets and result variables, so that it can be used by other projects that are not O3DE such as other 3p's, transparently.

Test Done:
- Tests of linking/compiling passed
- Tested with MultiplayerSample and verified the DTLS connection


package-system/OpenSSL/FindOpenSSL.cmake.template
```diff
@@ -6,29 +6,64 @@
...
+# we're trying to be a drop-in replacement for the FindOpenSSL.cmake that is shipped
+# with CMake itself, so we set the same variables with the same uppercase for compatibility
+# for questions about these variables, see https://cmake.org/cmake/help/latest/module/FindOpenSSL.html
+set(OPENSSL_FOUND True)
 set(OPENSSL_INCLUDE_DIR $${CMAKE_CURRENT_LIST_DIR}/OpenSSL/include)
 set(OPENSSL_LIBS_DIR $${CMAKE_CURRENT_LIST_DIR}/OpenSSL/$$<$$<CONFIG:debug>:debug/>lib)
+set(OPENSSL_CRYPTO_LIBRARY $${OPENSSL_LIBS_DIR}/libcrypto$${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(OPENSSL_CRYPTO_LIBRARIES
+    $${OPENSSL_CRYPTO_LIBRARY}
+    ${CRYPTO_LIBRARY_DEPENDENCIES})
```
spham-amzn, line 26: Did ${CRYPTO_LIBRARY_DEPENDENCIES} change from '${CUSTOM_ADDITIONAL_LIBRARIES}'? If so, I don't see the name change in build_package_image.py:

```python
builder.writeCMakeFindFile(
            outputDir,
            template=cmakeFindFileTemplate,
            templateEnv={
                'CUSTOM_ADDITIONAL_LIBRARIES':extraLibs,
                'CUSTOM_ADDITIONAL_COMPILE_DEFINITIONS':compileDefs,
            },
        )
```

package-system/OpenSSL/set_openssl_port_to_1_1_1_m.patch
```diff
@@ -1,70 +1,63 @@
-From 66c310b442ac237b4b7e8b733eea83e94621809f Mon Sep 17 00:00:00 2001
-Date: Thu, 15 Apr 2021 09:02:46 -0700
-Subject: [PATCH] Set OpenSSL version to 1.1.1b
+From 18884b079e3945b29e4adc3b33a76c19929f9b0a Mon Sep 17 00:00:00 2001
+Date: Tue, 8 Mar 2022 12:31:34 -0800
+Subject: [PATCH] Set OpenSSL version to 1.1.1m
```
lumberyard-employee-dm, line 3: Why do we need to patch OpenSSL?

spham-amzn: This looks like a patch to vcpkg (which is used to download and build), not openssl itself

package_build_list_host_windows.json
```diff
@@ -39,7 +39,7 @@
         "OpenEXR-3.1.3-rev2-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/OpenEXR --platform-name Windows --package-root ../../package-system --clean",
         "OpenMesh-8.1-rev3-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/OpenMesh --platform-name Windows --package-root ../../package-system --clean",
         "OpenSSL-1.1.1b-rev1-android": "package-system/OpenSSL/build_package_image.py --platform-name android",
-        "OpenSSL-1.1.1b-rev2-windows": "package-system/OpenSSL/build_package_image.py",
+        "OpenSSL-1.1.1m-rev1-windows": "package-system/OpenSSL/build_package_image.py",
```
cgalvan, line 42: The Qt package depends on OpenSSL so its config will need to be updated as well:

`3p-package-source\package-system\Qt\build_config.json`

spham-amzn: Actually it may not if we are removing the QtWebEngine

package-system/OpenSSL/set_openssl_port_to_1_1_1_m.patch
junbo75: Yeah, this is a patch for vcpkg to build OpenSSL 1.1.1m.

package-system/OpenSSL/FindOpenSSL.cmake.template
junbo75: Yeah, you're right. Seems like I didn't commit that change. Will fix itt.

package_build_list_host_windows.json
junbo75: Yeah, we have plan to update Qt as well after the OpenSSL upgrade. If we're removing the QtWebEngine, we may not need to update Qt.

package-system/OpenSSL/set_openssl_port_to_1_1_1_m.patch
```diff
@@ -1,70 +1,63 @@
-From 66c310b442ac237b4b7e8b733eea83e94621809f Mon Sep 17 00:00:00 2001
-Date: Thu, 15 Apr 2021 09:02:46 -0700
-Subject: [PATCH] Set OpenSSL version to 1.1.1b
+From 18884b079e3945b29e4adc3b33a76c19929f9b0a Mon Sep 17 00:00:00 2001
```
allisaurus, line 1: does the date after the hash ("Mon Sep 17...") need to be updated as well, or is this autogenerated?

package_build_list_host_windows.json
spham-amzn: QtNetwork may still use it, but i'm looking into see if its possible to use other options if possible

package-system/OpenSSL/FindOpenSSL.cmake.template
```diff
@@ -6,29 +6,64 @@
...
-    OPENSSL_ENABLED
-${CUSTOM_ADDITIONAL_COMPILE_DEFINITIONS})
-
-set(OPENSSL_LIBRARY
-    "$${OPENSSL_LIBS_DIR}/libssl$${CMAKE_STATIC_LIBRARY_SUFFIX}"
-    "$${OPENSSL_LIBS_DIR}/libcrypto$${CMAKE_STATIC_LIBRARY_SUFFIX}"
-${CUSTOM_ADDITIONAL_LIBRARIES})
+add_library($${SSL_TARGETNAME} STATIC IMPORTED GLOBAL)
+set_target_properties($${SSL_TARGETNAME} PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C")
+set_target_properties($${SSL_TARGETNAME} PROPERTIES IMPORTED_LOCATION "$${OPENSSL_CRYPTO_LIBRARY}")
```
lawsonamzn, line 37: Why does the ssl imported target use the crypto library?
I would have expected the ssl imported target use the ssl library, and the crypto imported target to use the crypto library.

package-system/OpenSSL/FindOpenSSL.cmake.template
```diff
@@ -6,29 +6,64 @@
...
+set(OPENSSL_CRYPTO_LIBRARY $${OPENSSL_LIBS_DIR}/libcrypto$${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(OPENSSL_CRYPTO_LIBRARIES
+    $${OPENSSL_CRYPTO_LIBRARY}
+    ${CRYPTO_LIBRARY_DEPENDENCIES})
+set(OPENSSL_SSL_LIBRARY $${OPENSSL_LIBS_DIR}/libssl$${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(OPENSSL_SSL_LIBRARIES $${OPENSSL_SSL_LIBRARY})
+set(OPENSSL_LIBRARIES
+    $${OPENSSL_CRYPTO_LIBRARIES}
+    $${OPENSSL_SSL_LIBRARIES})
+set(OPENSSL_VERSION_STRING "1.1.1m")
```
lawsonamzn, line 32: according to the cmake document, this is "OPENSSL_VERSION", not "OPENSSL_VERSION_STRING"

package-system/OpenSSL/FindOpenSSL.cmake.template
```diff
@@ -6,29 +6,64 @@
...
+add_library($${CRYPTO_TARGETNAME} STATIC IMPORTED GLOBAL)
+set_target_properties($${CRYPTO_TARGETNAME} PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C")
+set_target_properties($${CRYPTO_TARGETNAME} PROPERTIES IMPORTED_LOCATION "$${OPENSSL_CRYPTO_LIBRARY}")
 
-ly_target_include_system_directories(TARGET $${TARGET_WITH_NAMESPACE} INTERFACE $${OPENSSL_INCLUDE_DIR})
+add_library($${OPENSSL_O3DE_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+target_link_libraries($${OPENSSL_O3DE_NAMESPACE} INTERFACE $${OPENSSL_LIBRARIES})
 
-target_link_libraries($${TARGET_WITH_NAMESPACE} INTERFACE $${OPENSSL_LIBRARY})
+set(OPENSSL_COMPILE_DEFINITIONS OPENSSL_ENABLED)
```
lawsonamzn, line 46: This doesn't appear in the CMake FindOpenSSL , do we need it?

package-system/OpenSSL/FindOpenSSL.cmake.template
```diff
@@ -6,29 +6,64 @@
...
+set_target_properties($${SSL_TARGETNAME} PROPERTIES IMPORTED_LOCATION "$${OPENSSL_CRYPTO_LIBRARY}")
 
-add_library($${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+add_library($${CRYPTO_TARGETNAME} STATIC IMPORTED GLOBAL)
+set_target_properties($${CRYPTO_TARGETNAME} PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C")
+set_target_properties($${CRYPTO_TARGETNAME} PROPERTIES IMPORTED_LOCATION "$${OPENSSL_CRYPTO_LIBRARY}")
 
-ly_target_include_system_directories(TARGET $${TARGET_WITH_NAMESPACE} INTERFACE $${OPENSSL_INCLUDE_DIR})
+add_library($${OPENSSL_O3DE_NAMESPACE} INTERFACE IMPORTED GLOBAL)
+target_link_libraries($${OPENSSL_O3DE_NAMESPACE} INTERFACE $${OPENSSL_LIBRARIES})
```
lawsonamzn, line 44: Its probably better to make the official targets the real ones ,and then create an _alias_ to define the open3dengine 3rdPArty:xxxx one.

To be more specific, according to the official FindOpenSSL.cmake from cmake, after this script runs we expect the following targets:
OpenSSL::SSL
OpenSSL::Crypto

Open3Dengine can add its own, ie,
3rdParty::OpenSSL or whatever it wants to call it.

I'd recommend making the normal cmake-expected stuff so that its compatibel with the intrerface in CMAke's official FindOpenSSL.cmake, and then after that, add aliases to those targets for O3DE usage, ie,
```
add_libary(${SSL_TARGETNAME} STATIC IMPORTED)
... define the SSL_TARGETNAME library
add_library$(CRYPTO_TARGETNAME} STATIC IMPORTED)
... define the CRYPTO library...

...
add_library(${OPENSSL_O3DE_NAMESPACE} ALIAS ${SSL_TARGETNAME})

Also do the crypto and SSL targets depend on each other?  If they do, decide which one depends on the aother and make sure it has a target_link_libraries.

**Review by lawsonamzn — CHANGES_REQUESTED** [review]
Its looking good so far, but needs some straightening out (see comments)

Its also a pity this uses vcpkg.  It seems like so much overkill for such a low-dependency package...

File: package-system/OpenSSL/set_openssl_port_to_1_1_1_m.patch (hunk 4 above)
Comment (junbo75) [reply package-system/OpenSSL/set_openssl_port_to_1_1_1_m.patch:1]: That is autogenerated from git.

File: package-system/OpenSSL/FindOpenSSL.cmake.template (hunk 6 above)
Comment (junbo75) [reply package-system/OpenSSL/FindOpenSSL.cmake.template:32]: Yeah, actually OPENSSL_VERSION is set on the next line: 
set(OPENSSL_VERSION $${OPENSSL_VERSION_STRING})

Seems like we don't need OPENSSL_VERSION_STRING here and I'll remove it.

File: package-system/OpenSSL/FindOpenSSL.cmake.template (hunk 7 above)
Comment (junbo75) [reply package-system/OpenSSL/FindOpenSSL.cmake.template:46]: Searched in our code base and I don't see this compile definition used anywhere. Will remove it.

File: package-system/OpenSSL/FindOpenSSL.cmake.template (hunk 8 above)
Comment (junbo75) [reply package-system/OpenSSL/FindOpenSSL.cmake.template:44]: Sure, I'll check whether crypto and SSL targets depend on each other.

The reason I don't use an alias for OPENSSL_O3DE_NAMESPACE is that our target is the combination of SSL_TARGETNAME and CRYPTO_TARGETNAME.

Comment (lawsonamzn) [reply package-system/OpenSSL/FindOpenSSL.cmake.template:44]: Then the best way to do it is probably something like
```
add_libary($${SSL_TARGETNAME} STATIC IMPORTED GLOBAL)
... define the SSL_TARGETNAME library
add_library($${CRYPTO_TARGETNAME} STATIC IMPORTED GLOBAL)
... define the CRYPTO library...
add_library($${OPENSSL_O3DE_NAMESPACE} INTERFACE IMPORTED GLOBAL)

# depend on the other 2 targets for the 3rdparty O3DE one.
target_link_libraries($${OPENSSL_O3DE_NAMESPACE} INTERFACE $${CRYPTO_TARGETNAME) $${SSL_TARGETNAME})
```
In other words, when working with _targets_ keep the dependencies to other targets where possible as opposed to raw libraries.

File: package-system/OpenSSL/FindOpenSSL.cmake.template [hunk 9]
```diff
@@ -6,29 +6,65 @@
...
+set(SSL_TARGETNAME "OpenSSL::SSL")
+set(CRYPTO_TARGETNAME "OpenSSL::Crypto")
+
+# we're trying to be a drop-in replacement for the FindOpenSSL.cmake that is shipped
+# with CMake itself, so we set the same variables with the same uppercase for compatibility
+# for questions about these variables, see https://cmake.org/cmake/help/latest/module/FindOpenSSL.html
+set(OPENSSL_FOUND True)
 set(OPENSSL_INCLUDE_DIR $${CMAKE_CURRENT_LIST_DIR}/OpenSSL/include)
-set(OPENSSL_LIBS_DIR $${CMAKE_CURRENT_LIST_DIR}/OpenSSL/$$<$$<CONFIG:debug>:debug/>lib)
+# c-only pakcages can be released containing only Release executables
```
lawsonamzn, line 22: Typo. 
```
# C-only Packages
        ~~~~~~~
```

lumberyard-employee-dm (approved):

lawsonamzn (commented):
I take it that linux, mac, android, ios, are forthcoming next?

package-system/OpenSSL/FindOpenSSL.cmake.template
junbo75: Yes, I'm working on the upgrade for other platforms right now. I can keep this PR open and add any additional changes for other platform to it for review if that helps to save us some effort and time.

spham-amzn (approved):

lawsonamzn (approved):
This is great.  Next time we update it though we probably need to move it off vcpkg.

Jackie9527:
According to #95 , The Python package depends on OpenSSL.

junbo75:
@Jackie9527 I just checked with @spham-amzn and the Python package has already been updated to 1.1.1.m from the last security update https://github.com/o3de/3p-package-source/pull/95.There are built in scripts that is part of python that handles the dependencies
 and they have their own repo for 3rd party libs. There's no more update we need to make for Python. Please check the following lines for more details:
- https://github.com/o3de/3p-package-source/blob/2eb975f2ef03c7148c3b6f4b0e15c13537e57472/package-system/python/win_x64/open3d_python.patch#L93
- https://github.com/o3de/3p-package-source/blob/2eb975f2ef03c7148c3b6f4b0e15c13537e57472/package-system/python/win_x64/open3d_python.patch#L106

PR #103, opened 2022-03-15
libpng was shipping on linux and android in 'debug'

lawsonamzn:
The previous build scripts were using a non-multi-config generator and not specifying the CMAKE_BUILD_TYPE which leaves it blank and that often defaults to no-config or debug. 

This updates the build scripts to work on those platforms, strips out some unnecessary files, and fixes issues/bugs in the pull_or_build_from_git.py script that were blocking this script from functioning as expected.

Specifically, pull_and_build_from_git was using only the folder name when trying to remove it, and was also only printing out the folder name, which is pretty scary when in a build script thats rm-rfing
It was also not cleaning up its temp stuff between runs if --clean was specified, leading to weird situations where stale files are present despite --clean
It was also not correctly configuring IOS - changed the platform detection and toolchain files to declare the appropriate variables as per Apple's own open source cmake file.

This change also adds basic tests to make sure that the produced png package is linkable and can compile.


lawsonamzn:
I'm thinking its missing the NEON optimizations...
Oof.

amzn-sj (approved):

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -564,12 +564,18 @@ def build_and_install_cmake(self):
...
-            if not install_target_folder.is_dir():
-                install_target_folder.mkdir(parents=True)
         else:
             # Otherwise install directly to the target
             install_target_folder = self.build_install_folder
 
+        install_target_folder = install_target_folder.resolve().absolute()
+
+        if self.clean_build and install_target_folder.is_dir():
+            delete_folder(install_target_folder.absolute().resolve())
```
AMZN-alexpete, line 574: do we need the absolute().resolve()if we already did that on line 571?

lumberyard-employee-dm: The [Path.resolve](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) already makes the path absolute

package-system/libpng/test/test_png.c
```diff
@@ -0,0 +1,25 @@
...
+{
+    if (png_get_header_ver(0))
+    {
+        printf("png header version: %s\n", png_get_header_ver(0));
+        printf("test_png: All is OK!\n");
+        return 0;
+    }
+    printf("Failed to call png_get_header_ver(), error!\n");
+    return 1;
+}
```
AMZN-alexpete, line 25: nit: new line for linux

lumberyard-employee-dm (approved):

package-system/libpng/build_config.json
```diff
@@ -1,67 +1,113 @@
 {
     "git_url": "https://git.code.sf.net/p/libpng/code",
     "git_tag": "v1.6.37",
-    "package_name": "libpng",
-    "package_version": "1.6.37-rev1",
+    "package_name": "png",
```
AMZN-alexpete, line 4: Why the change in package name from `libpng` to `png`?

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -564,12 +564,18 @@ def build_and_install_cmake(self):
...
+
+        if not install_target_folder.is_dir():
+            install_target_folder.mkdir(parents=True)
```
spham-amzn, line 577: mdkir has an argument `exist_ok` which you can set to true that would remove the need for the check. However, there should be a check for `install_target_folder.is_file()` instead

```suggestion
        if install_target_folder.is_file():
            raise BuildError(f'Error: Target folder {install_target_folder} is a file')
        install_target_folder.mkdir(parents=True, exist_ok=True)
```

package-system/libpng/build_config.json
lawsonamzn: "png" is the actual official package name from CMake ( https://cmake.org/cmake/help/latest/module/FindPNG.html ) so that find_package(PNG) works.  I wanted to standardize in such that find_package(x) is serviced by package x-version-rev-platform to simplify things.

I could have made it "PNG" but most of our packages (if not all?) are lowercase to avoid confusion.    Note that the cmake target is still PNG and I will be updating o3de side to change from 3rdParty::libpng to 3rdParty::PNG to conform to cmake's built-in findPng, so it can be a drop in replacement for other 3p libs like OpenImageIO

AMZN-alexpete (approved):
Just a couple minor things.

lawsonamzn:
I cleaned up the various calls to creating and removing folders so they all work the same way on resolved paths and all check errors.

cgalvan (approved):

AMZN-alexpete (approved):
Nice!

PR #105, opened 2022-03-24
Openexr mac minos version fix

lawsonamzn:
Mac OS wasn't properly ingesting its toolchain file, so setting the "Minimum os supported" (Minos) was not working.  

As part of fixing this, I adjusted the repeated code to be a utility function instead, as well as added some more output logging for when tests begin and end.

To make sure that this works even in ninja multi-config mode, even with non-standard configuration types, on linux (Where a bug was reported) I added tests for this.

This also fixes a bug that occurs if you try to use the package downloader script but do not have optional Boto3 installed.  In that case, it will still work now, but just not use any boto3 classes or functionality


Scripts/cmake/o3de_package_utilities.cmake
```diff
@@ -0,0 +1,134 @@
...
+# pull_and_build_from_git.py (recommended):
+
+# "extra_files_to_copy": [
+#    ["../../Scripts/cmake/o3de_package_utilities.cmake", "o3de_package_utilities.cmake"]
+# ],
+
+# then use include(${CMAKE_CURRENT_LIST_DIR}/o3de_package_utilities.cmake) in your package Find file.
+
+# o3de_import_targets(
+    # NAMESPACE_FROM from_namespac]   <--- optional
```
hershey5045, line 22: typo: from_namespace

lawsonamzn: nicely spotted

Scripts/extras/package_downloader.py
```diff
@@ -31,7 +33,9 @@
     from botocore.exceptions import BotoCoreError
     from botocore.exceptions import ClientError
     _aws_s3_available = True
+    possible_download_errors = (ClientError, BotoCoreError, ssl.SSLError, URLError, OSError)
```
cgalvan, line 36: Nitpick: Could make this just `possible_download_errors = possible_download_errors + (ClientError, BotoCoreError)` in case you ever need to update the initial tuple

cgalvan (approved):

Scripts/cmake/o3de_package_utilities.cmake
```diff
@@ -0,0 +1,134 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+include(CMakeParseArguments)
```
lumberyard-employee-dm, line 9: CMakeParseArguments is not needed since [CMake 3.04](https://cmake.org/cmake/help/latest/module/CMakeParseArguments.html?highlight=cmakeparsearguments)

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -619,8 +621,14 @@ def build_and_install_cmake(self):
                                       '-S', str(cmakelists_folder.absolute()),
                                       '-B', str(self.build_folder.name)]
 
-                if self.custom_toolchain_file:
-                    cmake_generator_args.append( f'-DCMAKE_TOOLCHAIN_FILE="{self.custom_toolchain_file}"')
+                if self.package_info.custom_toolchain_file:
+                    custom_toolchain_file = self.package_info.custom_toolchain_file
+                    custom_toolchain_file_path = pathlib.Path(custom_toolchain_file).absolute().resolve()
```
lumberyard-employee-dm, line 626: The [resolve()](https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.Path.resolve) method makes the path absolute. 
Therefore the `absolute()` call is redundant

package-system/OpenEXR/build_config.json
```diff
@@ -2,35 +2,64 @@
...
-               "-DBUILD_SHARED_LIBS=OFF",
-               "-DCMAKE_INSTALL_PREFIX=\"../../openexr-windows/OpenEXR\""
+               "\"Visual Studio 16 2019\""
             ],
             "custom_test_cmd" : [
                "test_openexr_windows.cmd"
             ]
-         }
+         },
+         "Android": {
```
lumberyard-employee-dm, line 48: It's strange that we have an Android and iOS OpenEXR package when the library is only used with host tool applications.

Scripts/cmake/o3de_package_utilities.cmake
```diff
@@ -0,0 +1,134 @@
...
+            if (COMMAND ly_target_include_system_directories)
+                ly_target_include_system_directories(TARGET ${component_target_name} INTERFACE ${system_includes})
+            else()
+                target_include_directories(${component_target_name} SYSTEM INTERFACE ${system_includes})
+            endif()
+
+            # Alias the target in the new namespace:
+            if (to_namespace_prefix AND from_namespace_prefix AND NOT "${to_namespace_prefix}" STREQUAL "${from_namespace_prefix}")
+                add_library(${to_namespace_prefix}${component} ALIAS ${component_target_name})
+            endif()
```
lumberyard-employee-dm, line 83: What if the `from_namespace_prefix` is empty?
In that case an alias target will not be made.

Such as importing a target such as "3rdPartyWithoutNamespace" -> "3rdParty::3rdPartyWithoutNamespace"

package-system/OpenEXR/build_config.json
```diff
@@ -2,35 +2,64 @@
...
+      "-DOPENEXR_RUN_FUZZ_TESTS=OFF",
+      "-DBUILD_SHARED_LIBS=OFF",
+      "-DDOCS=OFF",
+      "-DPYTHON=OFF"
    ],
    "Platforms":{
       "Windows":{
          "Windows":{
-            "build_configs" : ["debug", "release"],
+            "build_configs" : ["Debug", "Release"],
```
hershey5045, line 36: Just wondering, does the case not matter? I noticed in visual studio we are using lower-case.

hershey5045 (approved):

package-system/OpenEXR/build_config.json
lawsonamzn: I went to the bother of creating and testing those configs (and they work).  But yes, I coul dremove them?  But then thats just more work if we ever want to support them.
Thoughts?

Scripts/cmake/o3de_package_utilities.cmake
lawsonamzn: I'll update that.  In the case of OpenEXR it won't make a difference, so I'm tempted to just keep the existing packages.

package-system/OpenEXR/build_config.json
lawsonamzn: I'll double check this, since I'm not sure why I changed it...  but bear in mind that these build configs are for the _package_ not for O3DE - as in, its the config to tell CMake to build the package source code that was downloaded from the official package site (like, from the OpenEXR homepage/git home).   Its very unlikely that for example OpenEXR comes with a (lowercase) "profile" or "debug" config, for example - it comes with the standard CMake ones (Release, Debug, RelWithDebInfo, and MinSizeRel).

Scripts/extras/pull_and_build_from_git.py
lawsonamzn: I'll remove it

Scripts/cmake/o3de_package_utilities.cmake
lawsonamzn: Weill remove.  Although its harmless, so I'm probably not going to rev the package itself for this.

Scripts/extras/package_downloader.py
lawsonamzn: sure!

package-system/OpenEXR/build_config.json
lumberyard-employee-dm: I think it is fine to keep it for now in the build config

lumberyard-employee-dm (approved):

AMZN-alexpete (approved):

package-system/OpenEXR/build_config.json
lawsonamzn: Yeah, this is correct, since its specifying the build configs to pass to cmake to build _the package itself_ from its original source code.  So it has to contain whatever the package expects (not what O3DE might expect)

cgalvan (approved):

lumberyard-employee-dm (approved):

lawsonamzn:
Thanks!  I'll get packages ready and push this

PR #110, opened 2022-04-01
[Mac] Fix the install id of the python library so that binaries that link to it will set the correct relative path to find the library.

amzn-sj:
We embed the whole Python.framework in app bundles that depend on Python(Editor, AP, o3de(ProjectManager)). Binaries that depend on it need to have the dependency specified as `@rpath/Python.framework/Versions/Current/Python` where the rpath would be set to something like `@executable_path/../Frameworks`. We currently fix this during our build when copying runtime dependencies(each build) [here](https://github.com/o3de/o3de/blob/8077789eeb4e921bba40add909c00e348b66bd93/cmake/Platform/Mac/runtime_dependencies_mac.cmake.in#L238). With this change we can remove that hack.

A new version of [Zlib](https://www.zlib.net/) was released on March 27th, 2022 and the old lib was moved to an archive link. So I updated the Zlib version.

Sphinx depends on Jinja2. But a recent update to Jinja2 deprecated imports(Sphinx issue: https://github.com/sphinx-doc/sphinx/issues/10291). This was fixed in Sphinx v4.0.0, but we can't update to that because that version also [removes](https://www.sphinx-doc.org/en/master/extdev/deprecated.html) `sphinx.domains.python.PyClassmember` and `sphinx.domains.python.PyModulelevel` which are still used by python [here](https://github.com/python/cpython/blob/74b95d86e0f14603f878c4df3133bc8a93f8f80a/Doc/tools/extensions/pyspecific.py#L38). We can't update Sphinx to a higher version so we limit Jinja2 to 2.11.3.

Jinja2 v2.3.11 uses `soft_unicode` which was removed in [MarkupSafe v2.1.0](https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0). This was fixed in Jinja v3.1.1. But we can't switch to that version because that would result in the previous issue.

Signed-off-by: amzn-sj <srikkant@amazon.com>


lumberyard-employee-dm (approved):

package-system/python/darwin_x64/open3d_python.patch
```diff
@@ -99,9 +99,9 @@ index 4fab488..2f1bd81 100755
 -                  ' -arch '.join(ARCHLIST),
 -              ),
 -          ),
-+              name="ZLib 1.2.11",
-+              url="https://zlib.net/zlib-1.2.11.tar.gz",
-+              checksum='1c9f62f0778697a09d36121ead88e08e',
++              name="ZLib 1.2.12",
```
spham-amzn, line 102: Bleeding edge.

PR #111, opened 2022-04-04
Update AWSNativeSDK for Android to use O3DE OpenSSL package

junbo75:
Signed-off-by: Junbo Liang <68558268+junbo75@users.noreply.github.com>

- Built Android AWS Native SDK with O3DE version of zlib and OpenSSL
- Verified with the [AWS Metrics gem test level](https://github.com/o3de/o3de/tree/development/AutomatedTesting/Levels/AWS/Metrics) on an Android device
- Note that this PR cannot be merged until the new version of Android OpenSSL is promoted


package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Android
```diff
@@ -14,14 +14,27 @@ set(AWS_BASE_PATH ${CMAKE_CURRENT_LIST_DIR}/${AWSNATIVESDK_PACKAGE_NAME})
...
 # Determine the lib path and any extra build dependencies
 if(LY_MONOLITHIC_GAME)
     set(AWSNATIVE_SDK_LIB_PATH ${AWS_BASE_PATH}/lib/$<IF:$<CONFIG:Debug>,Debug,Release>)
     set(AWSNATIVESDK_BUILD_DEPENDENCIES
         ${AWSNATIVE_SDK_LIB_PATH}/dependencies/${CMAKE_STATIC_LIBRARY_PREFIX}curl${CMAKE_STATIC_LIBRARY_SUFFIX}
-        ${AWSNATIVE_SDK_LIB_PATH}/dependencies/${CMAKE_STATIC_LIBRARY_PREFIX}ssl${CMAKE_STATIC_LIBRARY_SUFFIX}
-        ${AWSNATIVE_SDK_LIB_PATH}/dependencies/${CMAKE_STATIC_LIBRARY_PREFIX}crypto${CMAKE_STATIC_LIBRARY_SUFFIX}
-        ${AWSNATIVE_SDK_LIB_PATH}/dependencies/${CMAKE_STATIC_LIBRARY_PREFIX}z${CMAKE_STATIC_LIBRARY_SUFFIX}
+        OpenSSL::SSL
+        ZLIB::ZLIB
```
lawsonamzn, line 37: here's a couple observations here - 
1. Why would it only be a dependency if its monolithic?  SUrely these libraries would be used also in non-monolithic mode.
2. I notice that down below (see line number 65 and 73) we are not using IMPORTED_LOCATION to declare the main API library but instead relying on target_link_libraries.  This "works" but doesn't necessarily preserve the correct order.  This means that zlib or other dependencies could appear on the command line in a different order than expected, which doesn't matter to windows, but does matter to clang or gcc.   To do this properly, you'd set each library's IMPORTED_LOCATION or IMPORTED_LOCATION_DEBUG and so on to be the actual AWS library in question, then set target_link_libraries to be the dependencies of that library, so that everything comes in, in the right order.  If you need more information please let me know
4. This script will only work inside o3de.  It calls functions like ly_target_include_system_directories.   If this is the intention, that's fine, but o3de also automatically downlaods and unpacks and activates any packages that are in the format 3rdParty::xxxxx.  This means that those lines you have above here to find_package and so forth, are not actually necessary if this script will only run in o3de.  Instead you can depend on 3rdParty::ZLIB and 3rdParty:xxxx (whatever OpenSSLs 3rdParty alias is) and O3DE will automatically download it if it needs it.   The above block with teh IF NOT TRAGET part is just for if you want to use this find script outside o3de, like for example in a test application like libtiff and others have.

junbo75: 1. I'll check with @onecent1101 for these build dependencies. I didn't change the logic in the original cmake file but agree that it seems weird that we only need these dependencies for monolithic build.
2. Sure. I can look for some existing examples in the code base and reach out if I need more information on that.
3. Good to know. I'll make the changes to depend on 3rdParty::xxxxx.

junbo75: Synced offline. We have both static and shared AWS Native SDK libraries. Based on the current [cmake file](https://github.com/o3de/3p-package-source/blob/main/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Android), only the monolithic build depends on the static version of AWS Native SDK, and the other build type depends on the shared version of AWS Native SDK. The static zlib, openssl and curl libraries are built in the shared AWS Native SDK libraries so they only become build dependencies for the static AWS Native SDK libraries which is only used by monolithic build.

Addressed other comments in the next revision.

lawsonamzn (approved):

allisaurus (approved):

PR #112, opened 2022-04-20
OpenXR 3rdParty script

moraaar:
The script builds OpenXR library using github release tag `release-1.0.22`. It builds the library statically, in both debug and release configurations, on windows, android and linux.

The final packages generated only includes the headers, libraries and license file (which is Apache-2.0).

Tests done:
- Built locally the windows, linux and android packages successfully.
- Checked the content only includes headers, libraries and license file.
- Used the packages locally on O3DE, built correctly making Atom RHI depending on `3rdParty::OpenXR`.


lumberyard-employee-dm (approved):

amzn-phist (approved):

package-system/OpenXR/build_config.json
```diff
@@ -0,0 +1,65 @@
...
+      "-DOPENXR_DEBUG_POSTFIX=d"
+   ],
+   "cmake_install_filter":[
+      "*.h",
+      "*.lib",
+      "*.a"
+   ],
+   "build_configs": [
+      "Release",
+      "Debug"
```
spham-amzn, line 26: Are we still building debug builds for non-windows platforms? I thought we want to only have a debug configuration for windows only, and the rest only build the release version of the library?

spham-amzn: Mixing debug and non-debug is an issue with Windows for sure (for C++ libs). Yeah, there wasn't anything official about this, this came from a discussion in the 3PP slack channel. We didn't want to go back and change existing packages, but the consensus was that there was no real value in having debug 3P libraries unless they were necessary, which in the case of windows and c++ libraries they were (because of the dependence on windows release/debug c-runtime dll)

package_build_list_host_windows.json
```diff
@@ -54,6 +54,8 @@
         "tiff-4.2.0.15-rev4-android": "Scripts/extras/pull_and_build_from_git.py ../../package-system/tiff --platform-name Android --package-root ../../package-system --clean",
         "v-hacd-2.3-1a49edf-rev1-windows": "package-system/v-hacd/build_package_image.py --platform-name windows",
         "vulkan-validationlayers-1.2.198-rev1-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/vulkan-validationlayers --platform-name Windows --package-root ../../package-system --clean",
+        "OpenXR-1.0.22-rev1-android": "Scripts/extras/pull_and_build_from_git.py ../../package-system/OpenXR --platform-name Android --package-root ../../package-system --clean",
```
byrcolin, line 57: nit: the rest of the entries in here are in alphabetical order, so the OpenXR entries should go after the OpenSSL entries

byrcolin (approved):

package-system/OpenXR/build_config.json
moraaar: I see. Thanks for letting me know. Since this package is already created and tested I rather move on, keep it simple and same for all platforms. But I'll have it in mind the next time I create another 3rdParty package.

package_build_list_host_windows.json
moraaar: Done

spham-amzn (approved):

Adi-Amazon (approved):

PR #115, opened 2022-05-17
Openssl 1.1.1o

lawsonamzn:
This updates OpenSSL-1.1.1m to OpenSSL-1.1.1o
This solve some NVD vulnerabilities, as well as makes sure the minimum os version of ios and mac are adhered to.

Note that previously ios and mac tests were not running as part of the build.

Now they are.


package-system/OpenSSL/build_package_image.py
```diff
@@ -29,8 +31,9 @@ def main():
 
     packageSystemDir = Path(__file__).resolve().parents[1]
     opensslPackageSourceDir = packageSystemDir / 'OpenSSL'
-    outputDir = packageSystemDir / f'OpenSSL-{args.platformName}'
-    opensslPatch = opensslPackageSourceDir / 'set_openssl_port_to_1_1_1_m.patch'
+    outputDir = opensslPackageSourceDir / 'temp' / f'OpenSSL-{args.platformName}'
+
+    opensslPatch = opensslPackageSourceDir / 'set_openssl_port_to_1_1_1_x.patch'
```
lawsonamzn, line 36: Note that this was named _x intentionally since we could then reuse it for o, p, q, etc, without renaming a file and make diffs cleaner in the future.

package-system/OpenSSL/set_openssl_port_to_1_1_1_x.patch
```diff
@@ -0,0 +1,31 @@
...
+ 
+-set(OPENSSL_VERSION 1.1.1m)
++set(OPENSSL_VERSION 1.1.1o)
+ vcpkg_download_distfile(
+     ARCHIVE
+     URLS "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz" "https://www.openssl.org/source/old/1.1.1/openssl-${OPENSSL_VERSION}.tar.gz"
+     FILENAME "openssl-${OPENSSL_VERSION}.tar.gz"
+-    SHA512 ba0ef99b321546c13385966e4a607734df38b96f6ed45c4c67063a5f8d1482986855279797a6920d9f86c2ec31ce3e104dcc62c37328caacdd78aec59aa66156
++    SHA512 75b2f1499cb4640229eb6cd35d85cbff2e19db17b959ac4d04b60f1b395b73567f9003521452a0fcfeea9b31b26de0a7bccf476ecf9caae02298f3647cfb7e23
+ )
```
lawsonamzn, line 17: This is actually the part that makes vcpkg actually get the o version of the package instead of the m version.

spham-amzn (approved):

package-system/OpenSSL/set_openssl_port_to_1_1_1_x.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/ports/openssl/vcpkg.json b/ports/openssl/vcpkg.json
+index 516fc53..24f5696 100644
+--- a/ports/openssl/vcpkg.json
++++ b/ports/openssl/vcpkg.json
+@@ -1,6 +1,6 @@
+ {
+   "name": "openssl",
+-  "version-string": "1.1.1m",
++  "version-string": "1.1.1o",
+   "port-version": 2,
```
lawsonamzn, line 29: Wheras this just changes the display name and package name in vcpkg itself, not the actual zip it fetches.

package-system/OpenSSL/test_OpenSSL_android.cmd
```diff
@@ -11,13 +11,12 @@ mkdir temp\build_test
...
 @rem CMAKE demands forward slashes but PACKAGE_ROOT is in native path:
 set "PACKAGE_ROOT=%PACKAGE_ROOT:\=/%"
-set "DOWNLOADED_PACKAGE_FOLDERS=%DOWNLOADED_PACKAGE_FOLDERS:\=/%"
 
 cmake -S test -B temp/build_test ^
     -G Ninja ^
     -DCMAKE_BUILD_TYPE=Release ^
     -DCMAKE_TOOLCHAIN_FILE=../../../../Scripts/cmake/Platform/Android/Toolchain_android.cmake ^
-    -DCMAKE_MODULE_PATH="%DOWNLOADED_PACKAGE_FOLDERS%;%PACKAGE_ROOT%" || exit /b 1
+    -DCMAKE_MODULE_PATH="%PACKAGE_ROOT%" || exit /b 1
```
lawsonamzn, line 19: OpenSSL has no dependencies of its own, so it does not require downloaded packages.

lawsonamzn (commented):
Just leaving some explanatory comments myself as the author

lawsonamzn:
I do believe this is the last time we should use vcpkg for this.  After this we should rev to 3.0.x and also use the source code directly instead of vcpkg, it adds a lot of unnecessary complexity!

cgalvan (approved):

allisaurus (approved):

PR #119, opened 2022-05-25
OpenImageIO building on Windows/Linux/Darwin

cgalvan:
Update to OpenImageIO and OpenColorIO building on Windows, Linux, and Darwin:
- Updated the OIIO version to v2.3.12.0 instead of the latest (2.3.15.0) so that we don't have to update our OpenEXR version
- This file (https://github.com/OpenImageIO/oiio/blob/v2.3.12.0/.github/workflows/ci.yml) says that OIIO v2.3.12.0 is expected to be built with OCIO v2.1.1 (which I have bumped up to) and OpenEXR v3.1.3 (which is the version we are already building/packaging)
- Ensures the Python version running the build is `3.7.12` which is the same version of the dependency that we pull in, otherwise we won't be able to test the python bindings that get generated at the end
- Ensures `nasm` is installed on the build machine
- Builds debug versions of libraries for OpenImageIO, OpenColorIO, and boost
- Updated custom find files and distribution find files to work on all platforms and handle debug versions on windows
- Separated OIIO and OCIO executables and python bindings into separate targets `OpenImageIO::Tools::Binaries` and `OpenImageIO::Tools::PythonPlugins` (and same for OCIO) so that the consumer can pull in only what they actually need. The `OpenImageIO` and `OpenColorIO` targets just pull in the libs and headers
- Extended C++ test of built package to also test against OCIO and debug on windows
- Added test to verify that the python bindings for OIIO and OCIO work


package-system/OpenEXR/FindImath.cmake
```diff
@@ -14,13 +14,46 @@ include(${CMAKE_CURRENT_LIST_DIR}/o3de_package_utilities.cmake)
...
+        endforeach()
+    else()
+        message(WARNING "Target not found in Imath: ${component}")
+    endif()
+endforeach()
+
+# create main library alias that O3DE can use to get default dependencies.
+# users can also bind to any of the other components above, but this one gives you most of the functionality
+# you probably want:
+add_library(3rdParty::Imath ALIAS Imath::Imath)
```
lawsonamzn, line 56: Somethings not right here, this is a reversion to what it used to be?

package-system/OpenEXR/FindOpenEXR.cmake
```diff
@@ -24,6 +24,49 @@ if (NOT TARGET ZLIB::ZLIB)
...
+                                        MAP_IMPORTED_CONFIG_${UCONF} DEBUG)
+            else()
+                set_target_properties(${component} PROPERTIES 
+                                        MAP_IMPORTED_CONFIG_${UCONF} RELEASE)
+            endif()
+        endforeach()
+    else()
+        message(WARNING "Target not found in OpenEXR: ${component}")
+    endif()
+endforeach()
```
lawsonamzn, line 69: same here, this is discarding.  Perhasps something went wrong during conflict resolution?

package-system/OpenEXR/FindImath.cmake
cgalvan: Good catch, reverted!

package-system/OpenEXR/FindOpenEXR.cmake
cgalvan: Same, fixed!

lawsonamzn (approved):

AMZN-alexpete (approved):

lawsonamzn (approved):

PR #121, opened 2022-07-01
Improvements to pull_and_build_from_git

spham-amzn:
Some minor improvements and updates for pull_and_build_from_git

* Prevent unnecessary downloading of dependent 3P packages by evaluating the hash of the local package if any against whats to be downloaded
* Add an option to download packages (with hash) of additonal 3rd party tar balls for projects that need them

### DIFF
--- a/Scripts/extras/archive_downloader.py
+++ b/Scripts/extras/archive_downloader.py
@@ -124,7 +124,7 @@ def extract_package(src_package_file: str, target_folder:str):
     elif package_ext in ARCHIVE_EXTS_7ZIP:
         try:
             os.makedirs(destination_path, exist_ok=True)
-            subprocess.call(f"7z x -y {src_package_file_abs}", cwd=destination_path)
+            subprocess.call(f['7z', 'x', '-y', '{src_package_file_abs}'], cwd=destination_path)
         except Exception:
             raise RuntimeError(f"Archive file {src_package_file_abs} requires 7Zip to be installed and on the command path. ")
             
@@ -138,11 +138,11 @@ def extract_package(src_package_file: str, target_folder:str):
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description="Download, verify hash, and unpack remote zip file")
-    parser.add_argument('src_url', 
+    parser.add_argument('src-url', 
                         help='The download url of the zip package to download',
                         nargs=1)
     parser.add_argument('--hash-algorithm',
-                        help='The hash algorithm to use to calculate the fingerprint',
+                        help=f'The hash algorithm to use to calculate the fingerprint ({" ".join(SUPPORTED_HASH_ALGORITHMS.keys())}',
                         default='sha256',
                         required=False)
     parser.add_argument('--hash',
--- a/Scripts/extras/archive_downloader.py
+++ b/Scripts/extras/archive_downloader.py
@@ -9,11 +9,12 @@
 import argparse
 import hashlib
 import os
+import pathlib
 import platform
 import urllib.request
 import subprocess
-import zipfile
 import sys
+import zipfile
 
 SUPPORTED_HASH_ALGORITHMS = {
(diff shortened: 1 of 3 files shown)


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -571,8 +583,28 @@ def sync_source(self):
...
+                    print(f"Package {package_name} already downloaded")
+                else:
+                    if not PackageDownloader.DownloadAndUnpackPackage(package_name, package_hash, str(temp_packages_folder)):
+                        raise BuildError(f"Failed to download a required dependency: {package_name}")
+
+        # Check if there are any additional package dependencies to download and extract
+        if self.package_info.additional_download_packages:
+            print("Downloading additional packages")
+            for package_url, package_hash, package_algorithm in self.package_info.additional_download_packages:
+                print(f"Retrieving additional package {os.path.basename(package_url)} from {package_url}")
```
lumberyard-employee-dm, line 596: Seems like using basename on a url will only work if after the last slash the package name is available.

So for a link of something like `http://example.org/cdn/assimp-5.0.1-rev11-multiplatform.tar.xz` it would work, but it wouldn't work for some URL that points to a file like `http://example.org/cdn/RandomHash/ef67aa11204935`

Scripts/extras/archive_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+                        default='sha256',
+                        required=False)
+    parser.add_argument('--hash',
+                        help='The hash fingerprint to validate against',
+                        default='',
+                        required=False)
+
+    parser.add_argument('--target-folder',
+                        help='The target location for the download',
+                        required=True)
```
lumberyard-employee-dm, line 155: For consistency with the other options, can you make the positional argument use a dash("src-url") instead of an underscore("src_url")

Scripts/extras/archive_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Download, verify hash, and unpack remote zip file")
+    parser.add_argument('src_url', 
+                        help='The download url of the zip package to download',
+                        nargs=1)
+    parser.add_argument('--hash-algorithm',
+                        help='The hash algorithm to use to calculate the fingerprint',
```
lumberyard-employee-dm, line 145: Can you provide a list of example hash algorithms to use as part of help text for this argument?
Such as
```suggestion
                        help='The hash algorithm to use to calculate the fingerprint (sha256, md5, sha512, etc...)',
```

Scripts/extras/archive_downloader.py
```diff
@@ -0,0 +1,167 @@
...
+    elif package_ext in ARCHIVE_EXTS_TAR:
+        import tarfile
+        with tarfile.open(src_package_file_abs) as tar_file:
+            tarfile.extractall(destination_path)
+    elif package_ext in ARCHIVE_EXTS_7ZIP:
+        try:
+            os.makedirs(destination_path, exist_ok=True)
+            subprocess.call(f"7z x -y {src_package_file_abs}", cwd=destination_path)
+        except Exception:
+            raise RuntimeError(f"Archive file {src_package_file_abs} requires 7Zip to be installed and on the command path. ")
```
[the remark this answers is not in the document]
spham-amzn: This specific change is to support the pyside2 use case which needs a file that is only compressed with 7zip for windows:

https://download.qt.io/development_releases/prebuilt/libclang/libclang-release_130-based-windows-vs2019_64.7z

But I'll fix the subprocess call

Scripts/extras/pull_and_build_from_git.py
spham-amzn: Not sure if I follow, but the base name is purely informational, and is not really needed. I'll change to just printing the full package url instead

cgalvan (approved):

PR #122, opened 2022-07-28
3P Build Script Update for Python 3.10.5

spham-amzn:
Script updates to pull and build cpython version 3.10.5 from github

- Python 3.7.x will be end of life [Jun 2023](https://peps.python.org/pep-0537/). Python 3.10 is scheduled for end of life in [Oct 2026](https://peps.python.org/pep-0619/).
- Restore using locally built openssl due to the fact that ubuntu 22.04 has moved on to openssl3 and will not provide a system library version of openssl 1.1.1n
- A lot of the patching from 3.7.12 for previous security issues has been addressed already in 3.10.5. This change should help future security patches
- Part of changes for https://github.com/o3de/sig-platform/issues/54


package-system/python/darwin_x64/make-python.sh
```diff
@@ -201,22 +201,22 @@ echo "---------------- Copying Open3DEngine package metadata and license file --
...
+rm -f $SCRIPT_DIR/package/Python.framework/Versions/3.10/lib/python3.10/ensurepip/_bundled/pip-20*.whl
+cat $SCRIPT_DIR/package/Python.framework/Versions/3.10/lib/python3.10/ensurepip/__init__.py | sed 's/"20.1.1"/"22.0.3"/g' | sed 's/("pip", _PIP_VERSION, "py2.py3"),//g' > $SCRIPT_DIR/package/python/lib/python3.10/ensurepip/__init__.py_temp
+rm $SCRIPT_DIR/package/Python.framework/Versions/3.10/lib/python3.10/ensurepip/__init__.py
+mv $SCRIPT_DIR/package/Python.framework/Versions/3.10/lib/python3.10/ensurepip/__init__.py_temp $SCRIPT_DIR/package/python/lib/python3.10/ensurepip/__init__.py
 
 echo ""
 echo "----------------  Cleaning temp folder ----------------"
 echo ""
-rm -rf $SCRIPT_DIR/temp
+# rm -rf $SCRIPT_DIR/temp
```
lumberyard-employee-dm, line 219: So we don't clean up the temporary folder for building anymore?

spham-amzn: good catch, commented out for testing the patch, will restore

package-system/python/darwin_x64/open3d_python.patch
```diff
@@ -1,201 +1,22 @@
...
            dict(
--              name="Tcl 8.6.8",
--              url="ftp://ftp.tcl.tk/pub/tcl//tcl8_6/tcl8.6.8-src.tar.gz",
--              checksum='81656d3367af032e0ae6157eff134f89',
-+              name="Tcl 8.6.11",
-+              url="https://prdownloads.sourceforge.net/tcl/tcl8.6.12-src.tar.gz",
-+              checksum='87ea890821d2221f2ab5157bc5eb885f',
+               name="Tcl %s"%(tcl_tk_ver,),
+-              url="ftp://ftp.tcl.tk/pub/tcl//tcl8_6/tcl%s-src.tar.gz"%(tcl_tk_ver,),
++              url="https://fossies.org/linux/misc/tcl%s-src.tar.gz"%(tcl_tk_ver,),
```
lumberyard-employee-dm, line 10: This patch is for grabbing Tkinter python for Linux?
But this patch is in a MacOS specific file. Is this actually used?

spham-amzn: This is a source tarball, and i believe the linux/unix source also supports mac ( http://tcl.tk/software/tcltk/download.html )

package-system/python/win_x64/build_python.bat
```diff
@@ -29,7 +29,7 @@ set vswhere_location=%ProgramFiles(x86)%\Microsoft Visual Studio\Installer
 echo adding %vswhere_location% to PATH
 set PATH=%vswhere_location%;%PATH%
 
-for /f "tokens=*" %%i in ('vswhere -property installationPath') do set VS2017_LOCATION=%%i
+for /f "tokens=*" %%i in ('vswhere -version [15.0^,16.0^) -property installationPath') do set VS2017_LOCATION=%%i
```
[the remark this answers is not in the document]
spham-amzn: We don't, but it is still recommended to build cpython ( https://github.com/python/cpython/blob/3.10/PCbuild/readme.txt )

package-system/python/linux_x64/make-python.sh
```diff
@@ -72,60 +72,78 @@ cp -f -v libexpat/expat/lib/*.c cpython/Modules/expat/
...
 echo ""
-echo "--------------- Cloning bzip2 1.0.8 and building ---------------"
+echo "--------------- Cloning libffi 3.4.2 and building static version ---------------"
 echo ""
-git clone git://sourceware.org/git/bzip2.git --branch "bzip2-1.0.8" --depth 1
-if [[ ! -d "bzip2" ]]; then
-    echo "Was unable to create bzip2 dir via git clone.  Is git installed?"
+git clone https://github.com/libffi/libffi.git --branch "v3.4.2" --depth 1
+if [[ ! -d "libffi" ]]; then
+    echo "Was unable to create libffi dir via git clone."
```
lumberyard-employee-dm, line 79: We still need to build libfii?

lumberyard-employee-dm (approved):

package-system/python/linux_x64/make-python.sh
spham-amzn: Yes we do, that hasn't changed. Its passed into configure options for cpython below this line

cgalvan (approved):

package-system/python/darwin_x64/open3d_python.patch
```diff
@@ -1,201 +1,22 @@
...
-     os.rename(htmlDir, docdir)
-@@ -1522,6 +1524,11 @@ def buildInstaller():
-     pkgroot = os.path.join(outdir, 'Python.mpkg', 'Contents')
-     pkgcontents = os.path.join(pkgroot, 'Packages')
-     os.makedirs(pkgcontents)
-+
-+    #o3de modification:  We don't need packages for our purposes, so we return immediately:
-+    return
-+
-+
```
lawsonamzn, line 199: so this was all removed ... .. intentionally?  It works without this?  ANd it still correctly sets min OS to the right number?

lawsonamzn (commented):
You can check whether the "min os requirements" are correctly set on a binary using "otool -l"  Its one of the LD options, afaik (LC_VERSION_MIN_MACOSX).  It should be the same version as is set in the osx toolchain file in the 3p-package-scripts repo at Scripts/cmake/Platform/Mac

package-system/python/darwin_x64/open3d_python.patch
spham-amzn: None of the original patch from 3.7 applies to the 3.10 version of [build-installer.py](https://github.com/python/cpython/blob/v3.10.5/Mac/BuildScript/build-installer.py), but I wasn't notice that we were short-circuiting the packages.  Removing that short-circuit didn't seem to have any problems, all the python unit tests passed on Mac for me. I did not check the min OS number

spham-amzn: The original patch was partially restored (the part that dealt with the mac osx min version), and restored the skipping of the package creation. Verified with otool that it is set on the binary

```
otool -l python3 | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 10.15
      sdk 12.1
```

lawsonamzn: I guess 10.15 is okay, but the minos in our toolchain file is actually 11.0
we don't set a SDK version though (But perhaps thats related to minos?)

10.15 would work in all the situations we need it to.  But is it working by accident becuase we happen to be using that SDK to build it?

spham-amzn: 10.15 was specific to the MAc build scripts for python, even on the current version. https://github.com/python/cpython/blob/main/Mac/BuildScript/README.rst

lawsonamzn (approved):

lawsonamzn (approved):

Kadino:
These code changes look great, though I see that Automated Review is notably absent from the checks in the approval workflow for this repo.

Despite an expectation of backward compatibility, I would expect changing the interpreter version could cause some amount of dependency havoc in python scripts. How was this tested?

spham-amzn:
> These code changes look great, though I see that Automated Review is notably absent from the checks in the approval workflow for this repo.
> 
> Despite an expectation of backward compatibility, I would expect changing the interpreter version could cause some amount of dependency havoc in python scripts. How was this tested?

We dont have a process for that for the 3P scripts yet, but that is something that @amzn-changml is working on through a TDD process. For now, the changes are being tested through the ci_build.py testing scripts

Kadino (approved):

lawsonamzn:
> > These code changes look great, though I see that Automated Review is notably absent from the checks in the approval workflow for this repo.
> > Despite an expectation of backward compatibility, I would expect changing the interpreter version could cause some amount of dependency havoc in python scripts. How was this tested?
> 
> We dont have a process for that for the 3P scripts yet, but that is something that @amzn-changml is working on through a TDD process. For now, the changes are being tested through the ci_build.py testing scripts

Yes.  To be more clear though, updating a package here consists of 2 steps (or more).  Making the new build scripts for the package... but then updating AR on the O3DE side to use the new package.  if the new package does not function correctly, the change to O3DE that 'points it' at the new package will not pass AR on the O3DE side...  O3DE always uses a specific (verified by hash) package, so there's always a step where it goes thru the full AR before actually being used.

lawsonamzn (approved):

PR #124, opened 2022-08-11
3P Build Script Update for Pyside2 5.15.2.1

spham-amzn:
- Removed old pyside2-qt scripts which only represented the current Linux version, not the Windows version in favor of these scripts
- Updated the binary file structures to match the build from source with the original file structure rather than cherry picking them just the the site-packages link
- Updated to pull from pyside 5.15.2.1
- Added new 'pyside2::Tools' alias to support dependency on the shiboken tools that are part of the source build


package-system/pyside2/Findpyside2.cmake.Linux
```diff
@@ -0,0 +1,59 @@
...
+
+ly_pip_install_local_package_editable(${PYSIDE_BASE_PATH}/lib/python3.10/site-packages pyside2)
+
+set(${MY_NAME}_RUNTIME_DEPENDENCIES
+    ${PYSIDE_BASE_PATH}/lib/libpyside2.abi3.so.5.15.2.1
+    ${PYSIDE_BASE_PATH}/lib/libpyside2.abi3.so.5.15
+    ${PYSIDE_BASE_PATH}/lib/libpyside2.abi3.so
+    ${PYSIDE_BASE_PATH}/lib/libshiboken2.abi3.so.5.15.2.1
+    ${PYSIDE_BASE_PATH}/lib/libshiboken2.abi3.so.5.15
+    ${PYSIDE_BASE_PATH}/lib/libshiboken2.abi3.so
```
lumberyard-employee-dm, line 26: Seems like we should make two SHARED_IMPORTED GLOBAL targets for libpyside2 and libshiboken2, that is then added as a dependency of the pyside2 library

package-system/pyside2/build_linux.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Get the python executable from the package dependency
+LOCAL_PYTHON37_BIN=$TEMP_FOLDER/python-3.10.5-rev1-linux/python/bin/python3
+
+SCRIPT_PATH=`dirname $0`
+
+
+if [ ! -f "$LOCAL_PYTHON37_BIN" ]
+then
+    echo "Missing 3P dependency of python3.7 $LOCAL_PYTHON37_BIN"
```
lumberyard-employee-dm, line 20: This should be updated to mention that the dependent is python 3.10

package-system/pyside2/build_linux.sh
```diff
@@ -0,0 +1,75 @@
...
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+
+# Get the python executable from the package dependency
+LOCAL_PYTHON37_BIN=$TEMP_FOLDER/python-3.10.5-rev1-linux/python/bin/python3
```
lumberyard-employee-dm, line 13: We might want to change this variable to get rid of "37" from the name.

```suggestion
LOCAL_PYTHON3_BIN=$TEMP_FOLDER/python-3.10.5-rev1-linux/python/bin/python3
```

lumberyard-employee-dm (approved):

package-system/pyside2/build_linux.sh
spham-amzn: Done

package-system/pyside2/build_linux.sh
spham-amzn: Done

package-system/pyside2/Findpyside2.cmake.Linux
spham-amzn: (per offline discussion), the pyside2 target is used to specify dependent files that need to be copied to the runtime bin folder of the editor, and it is not linked against any code in O3DE. The binary files are needed by python when pyside2 is invoked through python, and having it reside in the same folder as the executable and the qt shared libraries ensures that the pyside site-packages and binaries link to the local o3de-specific Qt

sphrose (commented):
Just double checking: I can't see anything in the package_windows.bat script that's copying the shiboken generator exe. If it's being built to %INSTALL_SOURCE_RELEASE then that will cover it. 
There's also no explicit copying of typescript or plugin files - are they just covered under one of the other robocopies?

spham-amzn:
> Just double checking: I can't see anything in the package_windows.bat script that's copying the shiboken generator exe. If it's being built to %INSTALL_SOURCE_RELEASE then that will cover it. There's also no explicit copying of typescript or plugin files - are they just covered under one of the other robocopies?

Its done with robocopy here:

```
REM Copy the bin (release) folder 
mkdir %PACKAGE_BASE%\bin
robocopy %INSTALL_SOURCE_RELEASE%\bin %PACKAGE_BASE%\bin *.* /E
REM Copy over libclang
copy %TEMP_FOLDER%\libclang-release_130-based-windows-vs2019_64\libclang\bin\libclang.dll %PACKAGE_BASE%\bin\
```

sphrose (approved):

package-system/pyside2/Findpyside2.cmake.Windows
```diff
@@ -0,0 +1,47 @@
...
+    ${${MY_NAME}_BIN_DIR}/shiboken2${CMAKE_EXECUTABLE_SUFFIX}
+)
+
+set(${MY_NAME}_TOOLS_PYTHON_SCRIPTS
+    ${${MY_NAME}_BIN_DIR}/pyside_tool.py
+    ${${MY_NAME}_BIN_DIR}/shiboken_tool.py
+)
+
+set(${MY_NAME}_TOOLS_DEPENDENCIES
+    ${${MY_NAME}_BIN_DIR}/libclang.dll
```
lawsonamzn, line 34: How is this provided?  Do we ship this also?

package-system/pyside2/package_windows.bat
```diff
@@ -0,0 +1,120 @@
...
+copy %TEMP_FOLDER%\..\LICENSES.TXT %PACKAGE_BASE%\
+ECHO copy %TEMP_FOLDER%\src\README.* %PACKAGE_BASE%\
+copy %TEMP_FOLDER%\src\README.* %PACKAGE_BASE%\
+
+
+REM Copy the bin (release) folder 
+mkdir %PACKAGE_BASE%\bin
+robocopy %INSTALL_SOURCE_RELEASE%\bin %PACKAGE_BASE%\bin *.* /E
+REM Copy over libclang
+copy %TEMP_FOLDER%\libclang-release_130-based-windows-vs2019_64\libclang\bin\libclang.dll %PACKAGE_BASE%\bin\
```
lawsonamzn, line 42: I see we copy it here, but... where do we copy or provide the license file?  Is it a permitted license?

spham-amzn: Good catch. 
```
LLVM System Interface Library
-------------------------------------------------------------------------------
The LLVM System Interface Library is licensed under the Illinois Open Source
License and has the following additional copyright:

Copyright (C) 2004 eXtensible Systems, Inc.
```
I believe this refers to the University of Illinois Open Source License which is on the pre-approved list ( https://opensource.org/licenses/UoI-NCSA.php )

lawsonamzn: We should at least include that license, or put it in the same place as the files in the package, or etc. 
Basically we need to make sure everything is clear and disclosed.

spham-amzn: Will add that license file to the package

spham-amzn: The above license is now copied to LICENSE.LIBCLANG.TXT in the package

package-system/pyside2/build_linux.sh
```diff
@@ -0,0 +1,75 @@
...
+if [ ! -f "$LOCAL_PYTHON3_BIN" ]
+then
+    echo "Missing 3P dependency of python-3.10.5 $LOCAL_PYTHON3_BIN"
+    echo "You will need to build the 3P version of python under the '`readlink -f ../python`'' folder"
+    echo "with the following command:"
+    echo
+    echo "python3 build_package_image.py"
+    echo
+    exit 1
+fi
```
lawsonamzn, line 27: Is this still the case?  feels like the package downloader would have sorted this out.

lawsonamzn (approved):

package-system/pyside2/build_linux.sh
spham-amzn: This is due to the fact LOCAL_PYTHON3_BIN is hard-coded above, didn't want to add more to the script unless we have more cases where we use these package dependencies directly in the script

lawsonamzn (approved):

PR #126, opened 2022-08-27
Update AWSNativeSDK scripts to support building based on OpenSSL1 and 3

spham-amzn:
Updated 3P packaging to generate 2 Linux packages for AWS Native SDK:  One that links against OpenSSL 1.1.1, and the other that links against OpenSSL 3.0

* The script now uses Docker to setup the base environment based on specific distributions of Ubuntu that defaults to the specific version of OpenSSL: Ubuntu Focal Fossa (20.04) which uses OpenSSL 1.1.1 and Ubuntu Jammy Jellyfish (22.04) which uses OpenSSL  3.0
* The original scripts are now copied into the Docker image and ran inside there
* The resulting artifacts are copied from the Docker Container and packaged up as before
* The new patch that is applied allows for OpenSSL3 compatibility with the version of AWSNativeSDK that we use (1.9.50). The latest version of AWSNativeSDK as of this writing already has this patch, however there is a build error with it if you turn on CUSTOM_MEMORY_MANAGEMENT (see https://github.com/aws/aws-sdk-cpp/issues/1915 ). The patch file is generated from the commit that was done on aws's mainline ( https://github.com/aws/aws-sdk-cpp/commit/5f90c6e11c1e4a2b9234ecd2414a16de5bee0f4e )


lawsonamzn (approved):
Pretty cool!  Especially if the engine (on the other side) can figure it out based on what OSSL is available.  Perhaps find_package(openssl) will reveal it and it can use that result to pick the 1 or 3 series?  Is that the plan?

spham-amzn:
> Pretty cool! Especially if the engine (on the other side) can figure it out based on what OSSL is available. Perhaps find_package(openssl) will reveal it and it can use that result to pick the 1 or 3 series? Is that the plan?

It does, FindOpenSSL will set OPENSSL_MAJOR_VERSION as a cmake environment, and we will be using this to select the appropriate package(s)

package_build_list_host_linux.json
```diff
@@ -6,7 +6,8 @@
     "build_from_source": {
         "assimp-5.1.6-rev1-linux":  "Scripts/extras/pull_and_build_from_git.py ../../package-system/assimp --platform-name Linux --package-root ../../package-system --clean",
         "AWSGameLiftServerSDK-3.4.2-rev1-linux": "package-system/AWSGameLiftServerSDK/build_package_image.py --platform-name linux",
-        "AWSNativeSDK-1.9.50-rev1-linux": "Scripts/extras/pull_and_build_from_git.py ../../package-system/AWSNativeSDK --platform-name Linux --package-root ../../package-system --clean",
+        "AWSNativeSDK-1.9.50-rev2-linux-openssl1": "Scripts/extras/pull_and_build_from_git.py ../../package-system/AWSNativeSDK --platform-name Linux-OpenSSL1 --package-root ../../package-system/AWSNativeSDK/temp --clean",
+        "AWSNativeSDK-1.9.50-rev2-linux-openssl3": "Scripts/extras/pull_and_build_from_git.py ../../package-system/AWSNativeSDK --platform-name Linux-OpenSSL3 --package-root ../../package-system/AWSNativeSDK/temp --clean",
```
lumberyard-employee-dm, line 10: The 'L' being next to '1' makes it hard to read.
Could you put a a dash after openssl and the version number?

package-system/AWSNativeSDK/Dockerfile.ubuntu.2004
```diff
@@ -0,0 +1,49 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+
+# This docker file uses ubuntu 22.04 as the base image so that the AWS Native C++ libraries will use OpenSSL 3 as the base
```
spham-amzn, line 8: Fix comment: 
```suggestion
# This docker file uses ubuntu 20.04 as the base image so that the AWS Native C++ libraries will use OpenSSL 3 as the base
```

amzn-changml (approved):
Nice work! Might be nice to use this as a template for other 3p builds

junbo75 (approved):
Thanks for the updates.

lumberyard-employee-dm (approved):

package-system/AWSNativeSDK/Dockerfile.ubuntu.2004
```diff
@@ -0,0 +1,49 @@
...
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+
+# This docker file uses ubuntu 20.04 as the base image so that the AWS Native C++ libraries will use OpenSSL 3 as the base
+# for its dependencies
+#
+
+FROM ubuntu:20.04
```
allisaurus, line 12: (discussed offline) since this is to be used with the AWS SDK, suggest using a public ECR repository for this and the other dockerfile base image.
This verified Ubuntu repo is probably the right one to pull from: https://gallery.ecr.aws/ubuntu/ubuntu

allisaurus (approved):

lumberyard-employee-dm (approved):

PR #165, opened 2022-12-06
PhysX 5.1.1 build scripts

moraaar:
Added build script for PhysX 5 that can generate the following packages:
- PhysX-5.1.1-rev1-windows
- PhysX-5.1.1-rev1-linux
- PhysX-5.1.1-rev1-linux-aarch64
- PhysX-5.1.1-rev1-mac
- PhysX-5.1.1-rev1-ios
- PhysX-5.1.1-rev1-android

Signed-off-by: moraaar <moraaar@amazon.com>


lawsonamzn (commented):
looks pretty good to me, but are you sure you don't want to write a quick build test?

some of the other platforms (see openexr, etc) make a quick test of the package just by compiling a hello world and linking to it.  

Its not necessary but it could save you a lot of heartache when you find out that it doesn't link in some platform.  with a test, you can find that out from the build farm or locally when you build the package.  Without that it can only show up when the package is already up and live and then o3de is built against it.

Note that doing so, would require changing the find script to check for the various o3de specifc functions, might not be worth it, you decide...

package_build_list_host_windows.json
```diff
@@ -41,8 +41,8 @@
         "OpenSSL-1.1.1o-rev1-windows": "package-system/OpenSSL/build_package_image.py",
         "OpenXR-1.0.22-rev1-android": "Scripts/extras/pull_and_build_from_git.py ../../package-system/OpenXR --platform-name Android --package-root ../../package-system --clean",
         "OpenXR-1.0.22-rev1-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/OpenXR --platform-name Windows --package-root ../../package-system --clean",
-        "PhysX-4.1.2.29882248-rev5-android": "package-system/PhysX/build_package_image.py --platform android",
-        "PhysX-4.1.2.29882248-rev5-windows": "package-system/PhysX/build_package_image.py --platform windows",
+        "PhysX-5.1.1-rev1-android": "package-system/PhysX5/build_package_image.py --platform android",
+        "PhysX-5.1.1-rev1-windows": "package-system/PhysX5/build_package_image.py --platform windows",
```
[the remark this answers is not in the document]
moraaar: Yeah, I was wondering about that, but everything else is using `package-system`. Is it ok to use `package\temp`?

package_build_list_host_windows.json
```diff
@@ -107,8 +107,8 @@
     "OpenSSL-1.1.1o-rev1-windows": "package-system/OpenSSL/temp/OpenSSL-windows",
     "OpenXR-1.0.22-rev1-android": "package-system/OpenXR-android",
     "OpenXR-1.0.22-rev1-windows": "package-system/OpenXR-windows",
-    "PhysX-4.1.2.29882248-rev5-android": "package-system/PhysX-android",
-    "PhysX-4.1.2.29882248-rev5-windows": "package-system/PhysX-windows",
+    "PhysX-5.1.1-rev1-android": "package-system/PhysX5-android",
+    "PhysX-5.1.1-rev1-windows": "package-system/PhysX5-windows",
```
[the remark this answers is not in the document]
moraaar: ohhh good, thanks!

package_build_list_host_windows.json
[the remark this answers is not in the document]
moraaar: Scratch that, I'll use openexr as guidance. Thanks.

package_build_list_host_windows.json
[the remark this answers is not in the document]
moraaar: @lawsonamzn openssl, openexr and lua seem to still be using `package-system` folder... and not `package` so it won't be ignored by git... unless I'm missing something.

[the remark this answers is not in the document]
moraaar: done.

moraaar:
> looks pretty good to me, but are you sure you don't want to write a quick build test?
> ...
> Note that doing so, would require changing the find script to check for the various o3de specifc functions, might not be worth it, you decide...

That's a good suggestion. At the moment I'll skip it because there are other people waiting on this and I'd have to learn how to do tests and rebuild everything in all platforms, which takes a really long time. With that said PhysX 5 is exactly the same libs and includes as PhysX 4, so at least it's not a brand new library where more things can go wrong. In any case, I can look into it in the future and learn to add tests without the pressure of submitting and when convenient I'll just include the tests and update the packages.

package_build_list_host_darwin.json
```diff
@@ -79,8 +79,8 @@
         "tiff-4.2.0.15-rev3-ios": "package-system/tiff-ios",
         "python-3.10.5-rev2-darwin": "package-system/python/darwin_x64/package",
         "asn1-0.9.27-rev2-ios": "package-system/asn1-ios",
-        "PhysX-4.1.2.29882248-rev5-mac": "package-system/PhysX-mac",
-        "PhysX-4.1.2.29882248-rev5-ios": "package-system/PhysX-ios",
+        "PhysX-5.1.1-rev1-mac": "package-system/PhysX5/temp/PhysX5-mac",
+        "PhysX-5.1.1-rev1-ios": "package-system/PhysX5/temp/PhysX5-ios",
```
spham-amzn, line 83: This is a replacement of PhysX (4)? If so, then the package-system/PhysX folder should be deleted as well. If not, then we should keep the 4.X entries here, especially if there is a legacy bug fix case.

moraaar: I thought about it and since building PhysX 4.x is not trivial at all I rather not completely lose its build scripts for the moment and give the users the ability to generate both PhysX 4 and 5.

lawsonamzn: Yeah, no reason to delete the old package scripts and so on.  they're tiny and keeping them around means able to rebuild the old packages in case we ever need to...

spham-amzn: I was leaning on keeping both entries (PhysX-4 and PhysX-5)

moraaar: > I was leaning on keeping both entries (PhysX-4 and PhysX-5)

Sure, there is no harm. I'll add them back.

moraaar: done

SergeyAMZN (approved):

PR #175, opened 2023-01-26
Opencolorio/openimageio windows linkage fix

lawsonamzn:
Windows only problem :

The FindOpenColorIO/FindOpenImageIO files were linking to the release version even in debug, causing crashes.

I also discovered that the build script builds release versions of the standalone tool executables as well as python plugins.

To band-aid this, I made it so that it also copies the release dlls on windows.

At some point we need to go back and fix that, as well as fix our python package and rev it so that it has debug python in its find file.  But for now, this is at least an improvement that stops the editor crashing and also makes the command line tools usable.

fixes https://github.com/o3de/o3de/issues/14065


cgalvan (approved):

package-system/openimageio-opencolorio/test/test_OpenImageIO.cpp
```diff
@@ -60,11 +61,97 @@ int main()
...
+    if (!testWritingImage())
+    {
+        return 1;
+    }
+
+    if (!testWritingImage_tif_float())
+    {
+        return 1;
+    }
+
```
mbalfour-amzn, line 153: Shouldn't the tests clean up any temp images they write?

package-system/openimageio-opencolorio/test/.gitignore
```diff
@@ -0,0 +1,4 @@
+# test image files generated by our tests:
+temp_save_image*
+
+
```
mbalfour-amzn, line 4: Nit: This wouldn't be necessary if the test didn't leave artifacts around.

lawsonamzn: I'll fix these.  And amissing carriage return.  it doesn't seem like I need to publish a new package for these fixes, at least.

package-system/openimageio-opencolorio/test/test_OpenImageIO.cpp
mbalfour-amzn: Eh, yeah, fair point. But at least making it clean up in the common case would be nice. :)

lumberyard-employee-dm (approved):

mbalfour-amzn (approved):

PR #177, opened 2023-02-06
upgrade zlib version to 1.2.13, which is the latest release.

Jackie9527:
Signed-off-by: Jackie9527 <zhangqiangqiang@huawei.com>


lawsonamzn:
one thing to caution here, zlib is so base and fundamental to everything that this will potentially cascade to needing to then update many other packages.

Unless something really serious forces us to change zlib, I'd highly recommend against it.  Was it a security patch or something?

The situation here is that we distribute it as a static lib, so upgrading the package here will deliver a new static lib, and everything linking to that static lib will now link to the new one.   So you don't need to update other packages even if they depend on zlib.  However, if there are packages that deliver as dlls or executables, then those packages will pack teh contents of the static zlib into their own dlls and will have older versions.

If we're doing this for some new feature it probably doesn't matter.  if we're doing this to solve a real security hole, then we'd need to also issue new versions of the dll or exe distributed libraries in 3p that ingest the old zlib.

lawsonamzn (approved):
that being said, there's nothing I spy wrong with this. 
Its a pity the thing was one of the first packages ever made and didn't have any self tests tho

spham-amzn (approved):

KylerSF:
Just adding some info for anyone looking into this in the future, zlib 1.2.13 does fix a security problem, but o3de does not seem to use the affected piece of zlib from a quick search.
https://github.com/madler/zlib/releases/tag/v1.2.13
https://nvd.nist.gov/vuln/detail/CVE-2022-37434

PR #178, opened 2023-02-13
build sqlite lib linking against msvcrt.dll

Jackie9527:
Fix [7326](https://github.com/o3de/o3de/issues/7326),  [7188](https://github.com/o3de/o3de/issues/7188).
Link sqlite lib agains msvcrt, not libcmt, to remove the following warning: 
```
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library
```

Signed-off-by: Jackie9527 <80555200+Jackie9527@users.noreply.github.com>

### DIFF
--- a/package-system/sqlite/build_windows.cmd
+++ b/package-system/sqlite/build_windows.cmd
@@ -20,7 +20,7 @@ ECHO path as tcl is required to build for windows
 cd %BLD_PATH%
 
 REM Build the release library
-call nmake /f ..\src\Makefile.msc TOP=..\src libsqlite3.lib DEBUG=0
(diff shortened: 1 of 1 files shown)


lawsonamzn:
Nice!  The other linker warning we see is that it outputs pdbs but they are not included.  Is there a way to suppress PDB generation?  we don't really need to see the inside of sqlite or pay the cost of shipping them (which is why I bleieve they were stripped).  But a binary that thinks it made PDBs will beg for them vs a binary that never did...

lawsonamzn (approved):

spham-amzn (approved):

lumberyard-employee-dm (approved):

amzn-changml (approved):

PR #190, opened 2023-03-09
Fix missing include files for assimp package

spham-amzn:
The assimp package update to use docker and support for arm64 was missing the include header files from source. This fixes it by copying them over from the source into the build.
[build_assimp.log](https://github.com/o3de/3p-package-source/files/10928158/build_assimp.log)


lawsonamzn (approved):

lumberyard-employee-dm (approved):

cgalvan (approved):

spham-amzn:
Discovered that the current assimp libraries in prod were built with a different option than what was checked in, causing the libs to have undefined symbols during linking. Updating the build options to match whats in production.

[build_assimp.log](https://github.com/o3de/3p-package-source/files/10938167/build_assimp.log)

PR #192, opened 2023-03-18
Script updates for squish-ccr on Linux aarch64 (ARM64)

spham-amzn:
Updated linux build scripts for squish-ccr to support building for linux arm64. 

* For linux, switch to using Docker to build 
* Added ability to cross compile an aarch64 package on an x86_64 host using Docker + QEMU
* In order to support ARM and Neon, brought in an open-source header called ```sse2neon``` (with an permissive MIT license) to build the binary. This will not be part of the distribution but is used to compile the binaries for ARM
* Updated the injected CMakeList.txt to add handling of compiling to the armv8 architecture with neon
* Updated the existing patch to support using sse2neon

**Build Command**
```
 python3 ../3p-package-scripts/o3de_package_scripts/build_package.py --search_path . squish-ccr-deb557d-rev1-linux-aarch64
```

**Build output**
```

No tests defined, skipping test phase.
Creating/Updating package: squish-ccr-deb557d-rev1-linux-aarch64
    Adding files to: "./packages/temp/temp_package_squish-ccr-deb557d-rev1-linux-aarch64"
    Created: ./packages/squish-ccr-deb557d-rev1-linux-aarch64.tar.xz
    Created: ./packages/squish-ccr-deb557d-rev1-linux-aarch64.tar.xz.SHA256SUMS
    Created: ./packages/squish-ccr-deb557d-rev1-linux-aarch64.tar.xz.content.SHA256SUMS
    Created: ./packages/squish-ccr-deb557d-rev1-linux-aarch64.PackageInfo.json
    - Validating package: squish-ccr-deb557d-rev1-linux-aarch64 in folder ./packages....
    Package Hash: d3e54df2defff9f9254085acbf7c61dfda56f72ad10d34e1dd3b5d1bd2b8129f *squish-ccr-deb557d-rev1-linux-aarch64.tar.xz

```

**Upload output**
```
Output folder for packages is './packages' - Override with --output_folder
Not using an AWS profile.  Set LY_AWS_PROFILE or AWS_PROFILE or use command line params to change this.
Using bucket: ly-dev-3p-packages
Package: squish-ccr-deb557d-rev1-linux-aarch64 ...
    - Validating package: squish-ccr-deb557d-rev1-linux-aarch64 in folder ./packages....
    - Uploading squish-ccr-deb557d-rev1-linux-aarch64.tar.xz...
    - Uploading squish-ccr-deb557d-rev1-linux-aarch64.tar.xz.SHA256SUMS...
    - Uploading squish-ccr-deb557d-rev1-linux-aarch64.tar.xz.content.SHA256SUMS...
    - Uploading squish-ccr-deb557d-rev1-linux-aarch64.PackageInfo.json...
    - Uploaded package squish-ccr-deb557d-rev1-linux-aarch64.
```
[squish_ccr_build_log.txt](https://github.com/o3de/3p-package-source/files/11024417/squish_ccr_build_log.txt)


package-system/squish-ccr/CMakeLists.txt
```diff
@@ -107,11 +109,18 @@ set(SQUISH_PUBLIC_INCLUDE_FILES
...
-target_include_directories(squish-ccr PRIVATE ${CMAKE_CURRENT_LIST_DIR})
-target_compile_definitions(squish-ccr PRIVATE SQUISH_USE_SSE=2 SQUISH_USE_CPP SQUISH_USE_CCR)
+
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    target_compile_options(squish-ccr PRIVATE -msse2 -Wno-unused-value)
+    if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+        target_include_directories(squish-ccr PRIVATE ${CMAKE_CURRENT_LIST_DIR} /data/workspace/sse2neon)
+        target_compile_definitions(squish-ccr PRIVATE SQUISH_USE_NEON=1 SQUISH_USE_SSE=2 SQUISH_USE_CPP SQUISH_USE_CCR)
+        target_compile_options(squish-ccr PRIVATE -march=armv8-a+fp+simd+crypto+crc -Wno-unused-value -Wno-c++20-extensions -Wno-shift-op-parentheses -Wno-tautological-constant-out-of-range-compare -Wno-array-bounds -Wno-switch)
```
lumberyard-employee-dm, line 118: What is with this architecture?
`a+fp+simd+crypto+crc`?

What is `a`?

package-system/squish-ccr/Dockerfile.aarch64
```diff
@@ -0,0 +1,58 @@
...
+
+# Initilize apt cache
+RUN apt-get clean && apt-get update
+
+# Setup time zone and locale data (necessary for SSL and HTTPS packages)
+RUN DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata locales keyboard-configuration wget gpg
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
```
lumberyard-employee-dm, line 26: FYI wouldn't it be simpler to just append `en_US_UTF-8 UTF-8` to the end of the `locale.gen` file instead of attempting to replace a commented locale entry?
```suggestion
RUN echo 'en_US.UTF-8 UTF-8' >> /etc/locale.gen && \
    dpkg-reconfigure --frontend=noninteractive locales && \
    update-locale LANG=en_US.UTF-8
```

package-system/squish-ccr/Dockerfile.x86_64
```diff
@@ -0,0 +1,54 @@
+
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+
+# This docker file uses ubuntu 20.04 as the base image so that the AWS Native C++ libraries will use OpenSSL 3 as the base
+# for its dependencies
+#
```
lumberyard-employee-dm, line 10: The comment should be removed, since this docker file isn't for the AWS C++ SDKK

package-system/squish-ccr/Dockerfile.x86_64
```diff
@@ -0,0 +1,54 @@
...
+ 
+WORKDIR /data/workspace
+
+ARG DOCKER_BUILD_SCRIPT
+
+# Initilize apt cache
+RUN apt-get clean && apt-get update
+
+# Setup time zone and locale data (necessary for SSL and HTTPS packages)
+RUN DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata locales keyboard-configuration wget gpg
```
lumberyard-employee-dm, line 22: Why do we need support for the SSL and HTTPS packages in this docker?

package-system/squish-ccr/Findsquish-ccr.cmake
```diff
@@ -0,0 +1,45 @@
...
+set(LIB_NAME "squish-ccr")
+set(TARGET_WITH_NAMESPACE "3rdParty::${LIB_NAME}")
+if (TARGET ${TARGET_WITH_NAMESPACE})
+    return()
+endif()
+
+set(${LIB_NAME}_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/${LIB_NAME}/include)
+set(${LIB_NAME}_LIBRARY_DIR ${CMAKE_CURRENT_LIST_DIR}/${LIB_NAME}/bin)
+
+add_library(${TARGET_WITH_NAMESPACE} INTERFACE IMPORTED GLOBAL)
```
lumberyard-employee-dm, line 18: We can make this a STATIC IMPORTED GLOBAL target and avoid the use of target_link_libraries.
Instead we can use IMPORTED_LOCATION

package-system/squish-ccr/CMakeLists.txt
spham-amzn: Its actually 'armv8-a' for the cpu, followed by the options. That line is from the guide for sse2neon which is the project that is used to translate the sse calls to neon:
https://github.com/DLTcollab/sse2neon/blob/master/README.md

package-system/squish-ccr/Dockerfile.x86_64
spham-amzn: Its needed for the GPG keys for apt in the block below to get the latest cmake for ubuntu 20.04. The default version for 20.04 had issues with squish-ccr

package-system/squish-ccr/Dockerfile.aarch64
spham-amzn: I suppose, it wasn't an issue since this approach has been working on all the other docker images that I've created, and doing a `sed` seems as reasonable an approach as any

package-system/squish-ccr/CMakeLists.txt
lumberyard-employee-dm: Thanks for the info!
I was wondering how it combined together.

package-system/squish-ccr/Findsquish-ccr.cmake
spham-amzn: Does that apply even though this is a shared library?

lumberyard-employee-dm: For a SHARED library we should use SHARED IMPORTED GLOBAL then with the `IMPORTED_IMPLIB`  property on Windows only in addition to the `IMPORTED_LOCATION` property

lumberyard-employee-dm: We have an example of setting the properties for the OpenImageIO library: https://github.com/o3de/3p-package-source/blob/52e16f4c574eecf1f9fe02cd79bd48eebe75c38e/package-system/openimageio-opencolorio/distribution/FindOpenImageIO.cmake#L188-L196

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,135 @@
...
+# TEMP_FOLDER and TARGET_INSTALL_ROOT get set from the pull_and_build_from_git.py script
+
+if [ "$#" -eq 0 ]
+then
+    echo "Missing required [architecture] argument"
+    exit 1
+fi
+
+TARGET_AARCH=$1
+
```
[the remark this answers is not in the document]
spham-amzn: I can probably do additional updates to check for cross compilation here as well

lumberyard-employee-dm (approved):

package-system/squish-ccr/Findsquish-ccr.cmake
spham-amzn: Updated

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,172 @@
...
+elif [ "${TARGET_AARCH}" = "aarch64" ] 
+then
+    echo "Processing Docker for aarch64"
+    TARGET_DOCKER_FILE=Dockerfile.aarch64
+    TARGET_DOCKER_PLATFORM_ARG=linux/arm64/v8
+    DOCKER_IMAGE_NAME=squish_ccr_linux_aarch64_3p
+else
+    echo "Unsupported architecture ${TARGET_AARCH}"
+    exit 1
+fi
```
lumberyard-employee-dm, line 74: Instead of specifying `squish_ccr` 3 times throughout this file, how about adding a LIB_NAME variable that is set to "squish_ccr", so that it can be reused.

package-system/squish-ccr/build_config.json
```diff
@@ -55,24 +55,22 @@
...
-               "-j"
-            ],
-            "build_configs":[
-                "Release"
-            ]
-         }
-      }
+            "Linux":{
+               "custom_build_cmd": [
+                    "./build-linux-x86_64.sh"
```
spham-amzn, line 60: In order to reduce unnecessary scripts, i'll update the pull_and_build_from_git script to accept arguments to the 'custom_build_cmd' and adjust this accordingly

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -780,10 +781,20 @@ def build_and_install_custom(self):
         
         env_to_use = self.create_custom_env()
         custom_build_cmds = self.platform_config.get('custom_build_cmd', [])
-        for custom_build_cmd in custom_build_cmds:
+        if custom_build_cmds:
+
+            if len(custom_build_cmds)>1:
```
lumberyard-employee-dm, line 791: This could use some whitespace surrounding the `>`

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -792,11 +803,19 @@ def build_and_install_custom(self):
                 raise BuildError(f"Error executing custom build command {custom_build_cmd}")
 
         custom_install_cmds = self.platform_config.get('custom_install_cmd', [])
-       
-        for custom_install_cmd in custom_install_cmds:
+        if custom_install_cmds:
+
+            if len(custom_install_cmds)>1:
```
lumberyard-employee-dm, line 809: repeat: whitespace around the `>` sign would be nice.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -780,10 +781,20 @@ def build_and_install_custom(self):
...
+            else:
+                full_custom_build_cmd = custom_build_cmds[0].format(python=sys.executable)
+
+
+            print(f"CUSTOM BUILD COMMAND: {full_custom_build_cmd}")
+
             # Support the user specifying {python} in the custom_build_cmd to invoke
             # the Python executable that launched this build script
-            call_result = subprocess.run(custom_build_cmd.format(python=sys.executable),
+            call_result = subprocess.run(full_custom_build_cmd,
```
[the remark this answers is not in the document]
spham-amzn: In python2 it was recommended to pass in a single string when using `shell=True` ( https://docs.python.org/2/library/subprocess.html)
```
The shell argument (which defaults to False) specifies whether to use the shell as the program to execute. If shell is True, it is recommended to pass args as a string rather than as a sequence
```
I don't know if this is still the case for python3. I will check.

But I do see an alternative to this: https://docs.python.org/3/library/shlex.html

package-system/squish-ccr/docker_build_linux.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Failed to build squish-ccr for Linux (${build_config})"
+    exit 1
+fi
+
+cmake --install build/${build_config}
+if [ $? -ne 0 ]
+then
+    echo "Failed to package squish-ccr for Linux (${build_config})"
+    exit 1
+fi
```
lumberyard-employee-dm, line 31: This can be combined into a single call of `cmake --build build/${build_config} --target install`

package-system/squish-ccr/docker_build_linux.sh
```diff
@@ -0,0 +1,33 @@
...
+
+cmake --build build/${build_config}
+if [ $? -ne 0 ]
+then
+    echo "Failed to build squish-ccr for Linux (${build_config})"
+    exit 1
+fi
+
+cmake --install build/${build_config}
```
lumberyard-employee-dm, line 26: The build configs aren't being looped over

Also this can be simplified by using the "Ninja Multi-Config" generator

lumberyard-employee-dm: The following bash logic can be used to build both configurations
```bash
# Build only release variants
lib_name="squish-ccr"
build_configs=(Debug Release)

# Configure using the Ninja Multi-Config generator
cmake -S src -B build -G "Ninja Multi-Config" -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-fPIC -O2" -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=/data/workspace/package
if [ $? -ne 0 ]; then
    echo "Error configuring cmake for ${lib_name}"
    exit 1
fi

for config in "${build_configs[@]}"
do
    cmake --build build --target install
    if [ $? -ne 0 ]; then
        echo "Failed to build and package ${lib_name} for Linux (${config})"
        exit 1
    fi
done

exit 0

Comment (spham-amzn) [reply package-system/squish-ccr/docker_build_linux.sh:26]: we weren't / aren't building debug, just release.

File: package-system/squish-ccr/docker_build_linux.sh (hunk 12 above)
Comment (spham-amzn) [reply package-system/squish-ccr/docker_build_linux.sh]: works like a charm

File: package-system/squish-ccr/build-linux.sh [hunk 14]
```diff
@@ -0,0 +1,174 @@
...
+then
+    echo "Error: Cannot find Image ID for ${DOCKER_IMAGE_NAME}"
+    exit 1
+fi
+
+
+# Run the Docker Image
+echo "Running build script in the docker image"
+echo docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} -it --tty ${DOCKER_IMAGE_NAME}:latest /data/workspace/${DOCKER_BUILD_SCRIPT}
+docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} --tty ${DOCKER_IMAGE_NAME}:latest /data/workspace/${DOCKER_BUILD_SCRIPT}
```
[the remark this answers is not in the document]
spham-amzn: The problem is that the docker container runs as root, so the file permissions are set as root, not the local user that is running docker. That will cause issues with the `--clean` flag in the script, it won't be able to delete those root created artifacts without a `sudo`

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,174 @@
...
+if [ -z $IMAGE_ID ]
+then
+    echo "Error: Cannot find Image ID for ${DOCKER_IMAGE_NAME}"
+    exit 1
+fi
+
+
+# Run the Docker Image
+echo "Running build script in the docker image"
+echo docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} -it --tty ${DOCKER_IMAGE_NAME}:latest /data/workspace/${DOCKER_BUILD_SCRIPT}
```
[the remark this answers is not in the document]
spham-amzn: Oh, the echo was for troubleshooting purposes, and not really needed. `-it` enables you to run the docker container interactively, which is not needed by the script since it is invoking a script that is copied to the container. I will remove this line since the container is deleted at the end of this script anyways.

package-system/squish-ccr/Dockerfile.x86_64
```diff
@@ -0,0 +1,50 @@
...
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
+
+ENV LANG=en_US.UTF-8
+
+# Since we are using ubuntu 20.04, manually configure to get a newer version of cmake than the one that is the default for 20.04
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null && \
+    echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ focal main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null && \
```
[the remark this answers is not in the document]
spham-amzn: I don't think this is necessary, because this step was only needed for squish-ccr because ubuntu 20.04's default cmake was not compatible (i think). I think I will try and test it out without an updated cmake, which would eliminate the need for this block entirely.

package-system/squish-ccr/build-linux.sh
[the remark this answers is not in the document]
amzn-changml: Ah, ok. I was wondering if interactive input was actually needed here

package-system/squish-ccr/build-linux.sh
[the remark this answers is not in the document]
amzn-changml: Will using the `--user` flag work, or does it have to run the container as `sudo` after it's built? I didn't see if qemu needed `--privileged` so not sure if it need to need to run the build script as `root`?

package-system/squish-ccr/Dockerfile.x86_64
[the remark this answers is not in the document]
spham-amzn: Turns out this block (and others) wasn't necessary. The min version of cmake was arbitrarily set in this repo's version of CMakeList.txt, which may be a copy from O3DE's min version at the time. That version was changed to match ubuntu 20.04's default version instead.

package-system/squish-ccr/build-linux.sh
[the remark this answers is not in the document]
spham-amzn: Moved this out and will print it if an error occurs with a better explaination

package-system/squish-ccr/build-linux.sh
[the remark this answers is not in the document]
spham-amzn: The `--user` flag might work, i will give it a try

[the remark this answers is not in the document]
spham-amzn: It turns out `--user` applies to a user in the docker container, so in order for this to work, we would have to update the Dockerfile to also add the current user to the container, which seems more work than it is worth at this point.

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,184 @@
...
+# Copy over the license from sse2neon since it is used to build the aarch64 version of the library
+if [ "${TARGET_AARCH}" = "aarch64" ]
+then
+    docker cp --quiet $CONTAINER_ID:/data/workspace/sse2neon/LICENSE build/LICENSE.sse2neon
+    if [ $? -ne 0 ]
+    then
+        echo "Error occurred copying sse2neon license from Docker image ${DOCKER_IMAGE_NAME}:latest."
+        echo "To log into and troubleshoot the docker container, run the following command:"
+        echo ""
+        echo "docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} -it --tty ${DOCKER_IMAGE_NAME}:latest"
```
lumberyard-employee-dm, line 168: This isn't the correct docker command that ran.

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,184 @@
...
+
+# Run the Docker Image
+echo "Running build script in the docker image"
+docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} --tty ${DOCKER_IMAGE_NAME}:latest /data/workspace/${DOCKER_BUILD_SCRIPT}
+if [ $? -ne 0 ]
+then
+    echo Failed to build from docker image ${DOCKER_IMAGE_NAME}:latest
+    echo "To log into and troubleshoot the docker container, run the following command:"
+    echo ""
+    echo "docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} -it --tty ${DOCKER_IMAGE_NAME}:latest"
```
lumberyard-employee-dm, line 128: The `-it` option is still part of the echo statement.
```suggestion
    echo "docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} --tty ${DOCKER_IMAGE_NAME}:latest"
```

package-system/squish-ccr/build-linux.sh
```diff
@@ -0,0 +1,184 @@
...
+echo "Copying the built contents from the docker container for image ${DOCKER_IMAGE_NAME}"
+
+mkdir -p build
+docker cp --quiet $CONTAINER_ID:/data/workspace/package/. build  
+if [ $? -ne 0 ]
+then
+    echo "Error occurred copying build artifacts from Docker image ${DOCKER_IMAGE_NAME}:latest." 
+    echo "To log into and troubleshoot the docker container, run the following command:"
+    echo ""
+    echo "docker run --platform ${TARGET_DOCKER_PLATFORM_ARG} -it --tty ${DOCKER_IMAGE_NAME}:latest"
```
lumberyard-employee-dm, line 154: repeat: The echo'ed command is incorrect.

package-system/squish-ccr/build-linux.sh
spham-amzn: This was intentional. `-it` is not needed for the initial invocation because it runs a non-interactive script and will exit. Running with `-it` enables an interactive session, and it does not execute any specific script, this just launching into a shell.

package-system/squish-ccr/build-linux.sh
spham-amzn: See comment below

package-system/squish-ccr/build-linux.sh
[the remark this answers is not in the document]
amzn-changml: I believe there's a way to do it with just the UID (which you can get from `$(id -u $(logname))`), but I'm fine dropping this, as it's just an optimization.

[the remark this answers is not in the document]
spham-amzn: I did find the work-around, which uses the approach of mapping the user system folders:
```
CURRENT_UID=$(id -u)
CURRENT_GID=$(id -g)

# Run the Docker Image
echo "Running build script in the docker image"
docker run --user ${CURRENT_UID}:${CURRENT_GID} \
           -v $TEMP_FOLDER/src:/data/workspace/src \
           -v $TEMP_FOLDER/install:/data/workspace/install \
           -v "/etc/group:/etc/group:ro" \
           -v "/etc/passwd:/etc/passwd:ro" \
           -v "/etc/shadow:/etc/shadow:ro" \
           --platform ${TARGET_DOCKER_PLATFORM_ARG} \
           --tty \
           ${DOCKER_IMAGE_NAME}:latest /data/workspace/${DOCKER_BUILD_SCRIPT}
```
But it ended up also requiring changes to the docker script to chmod a+w on the workspace folder. In other words, it added more complexity to the script and the benefit didn't seem to justify it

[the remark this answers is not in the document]
amzn-changml: Fair enough. Dropping this

lumberyard-employee-dm (approved):

amzn-changml (approved):

PR #196, opened 2023-04-18
Updated AWSGameLiftServerSDK to new version 5.0.0

cgalvan:
Updated the `AWSGameLiftServerSDK` build script to fetch/build new version `5.0.0`.
NOTE: This currently only builds Windows, will be updating shortly for linux build.

Here is a summary of the changes:
- Updated `GAMELIFT_SERVER_SDK_DOWNLOAD_URL` to pull new `5.0.0` source zip
- New `5.0.0` version now depends on `OpenSSL` so added logic to retrieve our `OpenSSL` package on windows to use as a build dependency (linux uses the native OpenSSL so won't need to download it)
- Removed source patches as these were for external 3rd party dependencies that are no longer needed for the new version
- Removed libs of the aforementioned 3rd party dependencies from the find cmake files
- Added missing gitignore mentions for the `AWSGameLiftServerSDK` and `AWSNativeSDK` packages
- Ignored unzipping the `__MACOSX` folder from the source zip since we don't build `AWSGameLift` on darwin and it was causing max path length issues


lawsonamzn (approved):

AMZN-daimini (approved):

package_build_list_host_windows.json
```diff
@@ -7,7 +7,7 @@
         "assimp-5.2.5-rev1-windows":  "Scripts/extras/pull_and_build_from_git.py ../../package-system/assimp --platform-name Windows --package-root ../../package-system --clean",
         "astc-encoder-3.2-rev2-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/astc-encoder --platform-name Windows --package-root ../../package-system --clean",
         "azslc-1.8.15-rev2-windows": "Scripts/extras/pull_and_build_from_git.py ../../package-system/azslc --platform-name Windows --package-root ../../package-system/azslc/temp --clean",
-        "AWSGameLiftServerSDK-3.4.2-rev1-windows": "package-system/AWSGameLiftServerSDK/build_package_image.py --platform-name windows",
+        "AWSGameLiftServerSDK-5.0.0-rev1-windows": "package-system/AWSGameLiftServerSDK/build_package_image.py --platform-name windows",
```
allisaurus, line 10: will updating this entry (vs. adding a new one) affect users' ability to build the existing AWSGameLift gem, which currently relies on server SDK 3.4.2?

lumberyard-employee-dm: No, this will not affect building the AWSGameLift gem as that is using the GameLift Server SDK that is already uploaded to cloudfront.

The `package_build_list_host_<platform>.json` files here is used by the the 3p-package-scripts [build_package.py](https://github.com/o3de/3p-package-scripts#script-build_packagepy) script to build the actual 3rd Party from source.
That built library is then pushed to the production by sig-build once it is ready for use.
But the existing libraries remain in production.

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -155,13 +160,14 @@ def get_custom_build_env():
...
-def configure_sdk_project(source_folder: str,
+def configure_sdk_project(working_directory: WorkingDirectoryInfo,
                           build_folder: str,
                           build_type: str,
                           lib_type: str) -> None:
+    source_folder: str = working_directory.source_path.resolve()
     build_shared: str = "ON" if lib_type == "Shared" else "OFF"
     if PACKAGE_PLATFORM == PACKAGE_PLATFORM_OPTIONS[0]:
-        generator: str = "-G \"Visual Studio 15 2017\" -A x64"
+        generator: str = "-G \"Visual Studio 17 2022\""
```
lumberyard-employee-dm, line 170: The year isn't not needed in the generator.
"Visual Studio 17" works just as well in cmake
```suggestion
        generator: str = "-G \"Visual Studio 17\""
```

lumberyard-employee-dm (approved):

package_build_list_host_windows.json
cgalvan: Correct, Derric beat me to it :)

package-system/AWSGameLiftServerSDK/build_package_image.py
cgalvan: Sounds good I'll update that, thanks!

lawsonamzn (approved):

spham-amzn (commented):
Looks like updates to [package_build_list_host_*.json for linux, linux-aarch64, and mac are missing.

cgalvan:
> Looks like updates to [package_build_list_host_*.json for linux, linux-aarch64, and mac are missing.

I haven't built on linux yet I'm working on that now, and this package isn't built on mac.

PR #197, opened 2023-04-21
Updated AWSGameLiftServerSDK 5.0.0 to build on linux

cgalvan:
The `AWSGameLiftServerSDK` version `5.0.0` package only required a small change in order to build on linux (since we had already got it building on windows).

The server SDK tests are only built on linux, and they end up pulling down/building `googletest` as a dependency. There was actually a recent change that causes a compile error using `clang`:

https://github.com/google/googletest/issues/4206

However, I discovered in the gamelift README that `gcc` compiler is actually recommended over `clang`, and as mentioned in the bug, the Game Lift Server SDK builds fine with `gcc`, so the build script just needed to be modified to pass `gcc/g++` as the compiler instead.

### DIFF
--- a/package-system/AWSGameLiftServerSDK/build_package_image.py
+++ b/package-system/AWSGameLiftServerSDK/build_package_image.py
@@ -3,7 +3,7 @@
 #
 # Copyright (c) Contributors to the Open 3D Engine Project.
 # For complete copyright and license terms please see the LICENSE at the root of this distribution.
-# 
+#
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 #
 #
@@ -104,7 +104,7 @@ def collect_package_info() -> None:
         required=True,
         choices=PACKAGE_PLATFORM_OPTIONS
     )
-    
+
     args = parser.parse_args()
     global PACKAGE_PLATFORM
     PACKAGE_PLATFORM = args.platform_name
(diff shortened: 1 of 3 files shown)


cgalvan:
As discussed with @spham-amzn offline, we'll need to actually do 2 separate package builds for ubuntu 20.04 vs. 22.04, because they have different version of OpenSSL installed by default.

package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -152,8 +152,8 @@ def download_gamelift_server_sdk(working_directory: WorkingDirectoryInfo) -> Non
 def get_custom_build_env():
     if PACKAGE_PLATFORM in (PACKAGE_PLATFORM_OPTIONS[1], PACKAGE_PLATFORM_OPTIONS[2]):
         custom_env = os.environ.copy()
-        custom_env["CC"] = "clang"
-        custom_env["CXX"] = "clang++"
+        custom_env["CC"] = "gcc"
```
allisaurus, line 155: is `gcc` guaranteed to be present on systems where this script will run? And if not, will an error that indicates its absence be raised? (sorry if this is an obvious question, I am not super familiar with executing this project)

cgalvan: Yep! `gcc` comes by default in both our supported ubuntu versions (20.04 LTS and 22.04 LTS)

allisaurus (approved):

AMZN-Gene:
Does this version change inside [3p-package-source](https://github.com/o3de/3p-package-source) need to correspond with an o3de GameLift gem change?

AMZN-Gene (approved):

AMZN-daimini (approved):

cgalvan:
> Does this version change inside [3p-package-source](https://github.com/o3de/3p-package-source) need to correspond with an o3de GameLift gem change?

Yes, but this is just the build script, so it doesn't need to happen exactly in sync. Once the linux packages that have been built from the updated build script have been promoted to the production 3rd party bucket, then we can update the `package_build_list_host_linux.json` accordingly in the O3DE repo itself to pull down the new version.

PR #198, opened 2023-05-05
Updated windows build for GameLiftServerSDK to build against vs2019

cgalvan:
Updated the `GameLiftServerSDK ` build script to fix issue when linking against the library for monolithic builds. The `5.0.0-rev1` build was built against Visual Studio 2022, but if the user tries to build the AWSGameLift gem monolithically using Visual Studio 2019, they will run into two unresolved external symbol linker errors for missing `__std_init_once_link_alternate_names_and_abort` and `__std_find_trivial_1`, which are only available from the Visual Studio 2022 toolkit. Building with the Visual Studio 2019 toolkit will allow our users to build the gem on both 2019 and 2022.

Also while rebuilding discovered another windows max path being exceeded issue while unzipping the AWSGameLift-Go SDK from the release zip, so modified the build to only unzip the C++ SDK which is the only one we are building.


package-system/AWSGameLiftServerSDK/build_package_image.py
```diff
@@ -167,7 +169,10 @@ def configure_sdk_project(working_directory: WorkingDirectoryInfo,
     source_folder: str = working_directory.source_path.resolve()
     build_shared: str = "ON" if lib_type == "Shared" else "OFF"
     if PACKAGE_PLATFORM == PACKAGE_PLATFORM_OPTIONS[0]:
-        generator: str = "-G \"Visual Studio 17\""
+        # We need to explicitly build using the VS 2019 toolset. If we built using VS 2022
+        # then if we try to link against this library from VS 2019 there will be unresolved
+        # symbols for __std_init_once_link_alternate_names_and_abort and __std_find_trivial_1
+        generator: str = "-T v142"
```
allisaurus, line 175: it's not immediately apparent that `v142` = VS 2019 toolset. For my own understanding, why isn't `str = "-G \"Visual Studio 19\""` sufficient?

lumberyard-employee-dm: "Visual Studio 16" generator is only available if VS2019 is actually installed on the machine.
When using VS2022 to target an older toolset, the [-T v142 toolset](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html?highlight=cmake_generator_toolset) argument needs to be used to cmake()

allisaurus (approved):

AMZN-daimini (approved):

PR #202, opened 2023-05-23
Specified an explicit order of PhysX4 libraries

AMZN-Olex:
A specific order of PhysX4 libraries is required on Linux clang with monolithic builds of game project launchers.

The correct order was mentioned here https://github.com/NVIDIAGameWorks/PhysX/issues/53 and confirmed locally.

Tested the change locally by manually modifying the 3rdparty physx4 cmake script.


lawsonamzn (approved):
I think I previously investigated and found that indeed, the order specified here is preserved in the final linker line.

package-system/PhysX/FindPhysX.cmake.template
```diff
@@ -55,17 +55,38 @@ set(extra_static_libs ${EXTRA_STATIC_LIBS_NON_MONOLITHIC})
...
+            $${MONO_PATH_TO_STATIC_LIBS}/$${CMAKE_STATIC_LIBRARY_PREFIX}$${PHYSICS_LIB}$${CMAKE_STATIC_LIBRARY_SUFFIX}
         )
-    endif()
+        set (PREVIOUS_PHYSICS_LIB $${PHYSICS_LIB}::imported)
+    endforeach()
+
+    add_library(Physx4_STATIC_LIBS_FOR_MONOLITHIC::imported INTERFACE IMPORTED GLOBAL)
+    target_link_libraries(Physx4_STATIC_LIBS_FOR_MONOLITHIC::imported INTERFACE
+        PhysXFoundation_static_64::imported
+    )
```
AMZN-alexpete, line 89: does this change mean we are statically linking PhysX libs into the monolithic .exe and no longer need to include the PhysX runtime deps alongside the runtime?  If so, was there a legal/licensing reason we weren't doing that before?

AMZN-alexpete: Also, it looks like PhysX5 still uses the old code - does it also need this change?

AMZN-Olex: > does this change mean we are statically linking PhysX libs into the monolithic .exe and no longer need to include the PhysX runtime deps alongside the runtime? If so, was there a legal/licensing reason we weren't doing that before?

There is still PhysX GPU shared library but all others libraries are static.

> Also, it looks like PhysX5 still uses the old code - does it also need this change?

I haven't tested PhysX5 case. That is behind a switch case, so I would tackle it afterwards. @AMZN-alexpete

AMZN-alexpete (approved):

package-system/PhysX/FindPhysX.cmake.template
```diff
@@ -55,17 +55,38 @@ set(extra_static_libs ${EXTRA_STATIC_LIBS_NON_MONOLITHIC})
...
-        set($${MY_NAME}_RUNTIME_DEPENDENCIES
-            $${extra_shared_libs}
+    foreach(PHYSICS_LIB $${IMPORTED_PHYSICS_LIBS})
+        add_library($${PHYSICS_LIB}::imported STATIC IMPORTED GLOBAL)
+        set($${PHYSICS_LIB}_PATH $${MONO_PATH_TO_STATIC_LIBS}/$${CMAKE_STATIC_LIBRARY_PREFIX}$${PHYSICS_LIB}$${CMAKE_STATIC_LIBRARY_SUFFIX})
+        set_target_properties($${PHYSICS_LIB}::imported
+            PROPERTIES
+                IMPORTED_LOCATION_DEBUG   $${$${PHYSICS_LIB}_PATH}
+                IMPORTED_LOCATION_PROFILE $${$${PHYSICS_LIB}_PATH}
+                IMPORTED_LOCATION_RELEASE $${$${PHYSICS_LIB}_PATH}
```
lumberyard-employee-dm, line 77: Since we are only using the release library on for PhysX linux, only a single IMPORTED_LOCATION is needed
```suggestion
                IMPORTED_LOCATION   $${$${PHYSICS_LIB}_PATH}
```

package-system/PhysX/FindPhysX.cmake.template
```diff
@@ -55,17 +55,38 @@ set(extra_static_libs ${EXTRA_STATIC_LIBS_NON_MONOLITHIC})
...
+        )
+        target_link_libraries($${PHYSICS_LIB}::imported INTERFACE 
+            $${PREVIOUS_PHYSICS_LIB}
```
lumberyard-employee-dm, line 80: NOTE for others: This is the part which enforces the correct order.
Each successive physics library depends on the previous library.

lumberyard-employee-dm (approved):

AMZN-Olex:
Updated the PR with Physx5 changes.

AMZN-alexpete (approved):
Thanks for fixing PhysX5, super helpful for robotics sim dev!

nick-l-o3de (approved):

PR #205, opened 2023-07-24
Linux: Allow for conditional use of system SQLite library instead of bundled

nicholas-rh:
This is a first go at an implementation of the feature request outlined here: https://github.com/o3de/o3de/issues/16375

This PR allows for the user to specify to use the system sqlite library instead of the bundled statically linked version. The high-level goal here is to work towards enabling builds using the available system libraries where it makes sense and is useful. This would reduce the burden on package maintainers (and ideally the O3DE build systems/setup for Linux as well).

I tested building the install target both with the cmake variable enabled and disabled by building the sqlite package with the new find scripts included, installing them to my LY_3RDPARTY_PATH, and updating the package hash, and verified the expected behavior. 

The main area which may need some extra feedback/review is the impact on monolithic & distributable builds and if there needs to be some extra logic somewhere for those, since we are dynamically linking instead of statically, and truthfully I'm not as familiar with the monolithic build workflow.


nicholas-rh:
Also just to clarify I don't intend for this to replace the sqlite O3DE package, or even be the primary supported method of linking to sqlite, but just to exist as another option for those who find it useful to reduce the amount of cmake script hacking required

PR #209, opened 2023-07-28
Add cmake cache var to allow for use of system openssl

nicholas-rh:
See https://github.com/o3de/o3de/issues/16375 and https://github.com/o3de/3p-package-source/pull/205

This one was a bit trickier than SQLite, since the custom FindOpenSSL.cmake module takes precedent over the prewritten find module that is provided by cmake since they have the same name. SQLite doesn't have this problem because the system SQLite find module is "FindSQLite3.cmake" as opposed to the O3DE find module "FindSQLite.cmake," so there is no name collision. I ran into the same problem with the Freetype library, so it's not limited to just OpenSSL.

To get around this I manually set and then unset CMAKE_MODULE_PATH to an empty string. This workaround works because normal cmake variables have priority over cmake cache variables, and since CMAKE_MODULE_PATH is empty, the system find module is picked up as a fallback. 

The better solution would probably be to rename each find module to follow the pattern "FindO3DE_OpenSSL.cmake" or something similar and then do "find_package(O3DE_OpenSSL)" so there is no name collision, but that would be a bit more invasive and we would need to make sure to coordinate the changes in both this repo and the main o3de repo. If there's an easier solution I'm missing let me know.

### DIFF
--- a/package-system/OpenSSL/FindOpenSSL.cmake.Linux.template
+++ b/package-system/OpenSSL/FindOpenSSL.cmake.Linux.template
@@ -1,0 +1,91 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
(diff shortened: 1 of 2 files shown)


nicholas-rh:
I've modified this to store the value of CMAKE_MODULE_PATH in a temporarily variable to be restored after invoking find_package().

PR #214, opened 2023-09-14
Apply Fix for Qt running in Snap Packages

spham-amzn:
**Summary**
This fixes a [QT bug](https://bugreports.qt.io/browse/QTBUG-83939) when running the Project Manager from a Linux Snap package.

* Applying a patch similar to the [fix](https://codereview.qt-project.org/c/qt/qtbase/+/475629/2/src/gui/platform/unix/qgenericunixservices.cpp) onto the [o3de fork](https://github.com/o3de/qt5.git) of Qt when building from source.
* Fixing build config to conform to how the custom *scripts are evaluated
* Updating and applying a previous update to link against the static version of OpenSSL from O3DE's 3P repository as well.
* Update the linux revision numbers for the new fixed version

**Build Logs**
[qt_build_linux_rev9_x86_64.log](https://github.com/o3de/3p-package-source/files/12612325/qt_build_linux_rev9_x86_64.log)

[qt_build_linux_rev9_aarch64.log](https://github.com/o3de/3p-package-source/files/12612327/qt_build_linux_rev9_aarch64.log)

Fixes https://github.com/o3de/o3de/issues/16697


cgalvan (approved):
This works. Alternatively, we could just commit patches to our Qt fork which is why we originally created the forks (we had too many patches to keep track of).

spham-amzn:
> This works. Alternatively, we could just commit patches to our Qt fork which is why we originally created the forks (we had too many patches to keep track of).

Thats a good point, i'll create a GHI to do that in the future

cgalvan:
> > This works. Alternatively, we could just commit patches to our Qt fork which is why we originally created the forks (we had too many patches to keep track of).
> 
> Thats a good point, i created a [GHI](https://github.com/o3de/3p-package-source/issues/215) for that

Cool, thank you!

package-system/Qt/build-linux-x86.sh
```diff
@@ -11,7 +11,8 @@
 
 TIFF_PACKAGE=tiff-4.2.0.15-rev3-linux
 ZLIB_PACKAGE=zlib-1.2.11-rev5-linux
+OPENSSL_PACKAGE=OpenSSL-1.1.1t-rev1-linux
```
lumberyard-employee-dm, line 14: Is there a way that we can get these set as environment variables using the build_config.json, so that we don't have the package revision in scripts.?

spham-amzn: Not without more changes to `pull_and_build_from_git`. These packages are defined using the `depends_on_packages` attribute in the build_config.json, which would require the addition of an additional 'KEY' field, and then all projects will need to update using this key. I don't mind doing this, but i want to reduce the scope of this change to focus on the Qt fix. I can create another PR to support this and update all the current projects in the future.

lumberyard-employee-dm: That sounds good to me.

lumberyard-employee-dm (approved):

package-system/Qt/build-linux-x86.sh
spham-amzn: GHI: https://github.com/o3de/3p-package-source/issues/217

PR #229, opened 2023-11-28
Fix AWSNativeSDK Dependencies

AMZN-Gene:
Fix AWSNativeSDK library dependencies which would randomly cause compile errors depending on how CMake configure generated the linker ordering. 
All AWSNativeSDK libs depend on the Core library, but if lucky, the core lib was linked in before other libs (like Lamdba or S3)

Fixes https://github.com/o3de/o3de/issues/17072


package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux
```diff
@@ -121,6 +121,8 @@ ly_declare_aws_library(
         AccessManagement
     LIB_FILE 
         aws-cpp-sdk-access-management
+    BUILD_DEPENDENCIES
+        3rdParty::AWSNativeSDK::Core
```
AMZN-alexpete, line 125: it may be worth a comment in one or two of these files explaining why we are explicitly declaring this build dependency here to resolve an issue with order so someone doesn't remove it in the future.

AMZN-alexpete (approved):
Please leave some comments so future devs know why this explicit build dependency is here (to resolve issues with linker lib orders)

lumberyard-employee-dm (approved):

lumberyard-employee-dm:
> Please leave some comments so future devs know why this explicit build dependency is here (to resolve issues with linker lib orders)

Perhaps you should add a link to the 3p-package-source repo with a small blerb explaining about static library link order dependencies.
This documentation for the gcc `-l` linker option https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-Libraries explains that the order of linkage is important.
Also this stackoverflow answer might help future library maintainers as well https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc/409470#409470.

CMake will have static libraries linked in the proper order as long as they are _actually_ marked as STATIC libraries using the [add_library](https://cmake.org/cmake/help/latest/command/add_library.html) function.

There was problems early on with O3DE 3rdParty libraries where we we would mark an a 3rdParty library as an `INTERFACE` and then use the `target_link_libraries` command to add both the main file associated with the 3rdParty and then it's dependencies.
For example libtiff use to be marked as a INTERFACE library, where it used the `target_link_libraries` function to add the path to the libtiff.a file and a dependency on the Zlib library.

In that case the libtiff.a file was **NOT** dependent on Zlib, but instead just another listed in the same `target_link_libraries` command. Therefore when CMake generated a Makefile or Ninja ruile file it was free to re-order the listing of those libraries.

Pretty much the property way to declare a static library dependency can be seen in the [libtiff](https://github.com/o3de/3p-package-source/blob/7901137c1edf74d3f8c7c9fe1081e87cf29a4af2/package-system/tiff/FindTIFF.cmake#L50-L55) cmake today

#### Proper way to declare dependency: Good
```cmake
# add the CMake standard TIFF::TIFF library.  It is a static library.
add_library(TIFF::TIFF STATIC IMPORTED GLOBAL)
set_target_properties(TIFF::TIFF PROPERTIES IMPORTED_LOCATION "${TIFF_LIBRARY}")

target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB)
```

#### Improper way to declare dependency: Bad
```cmake
# add the CMake standard TIFF::TIFF library.  It is an interface library.
add_library(TIFF::TIFF INTERFACE IMPORTED GLOBAL)

target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB "${TIFF_LIBRARY}") # No actual dependency between ZLIB and TIFF and therefore can be reordered
```

AMZN-Gene:
> > Please leave some comments so future devs know why this explicit build dependency is here (to resolve issues with linker lib orders)
> 
> Perhaps you should add a link to the 3p-package-source repo with a small blerb explaining about static library link order dependencies. This documentation for the gcc `-l` linker option https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-Libraries explains that the order of linkage is important. Also this stackoverflow answer might help future library maintainers as well https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc/409470#409470.
> 
> CMake will have static libraries linked in the proper order as long as they are _actually_ marked as STATIC libraries using the [add_library](https://cmake.org/cmake/help/latest/command/add_library.html) function.
> 
> There was problems early on with O3DE 3rdParty libraries where we we would mark an a 3rdParty library as an `INTERFACE` and then use the `target_link_libraries` command to add both the main file associated with the 3rdParty and then it's dependencies. For example libtiff use to be marked as a INTERFACE library, where it used the `target_link_libraries` function to add the path to the libtiff.a file and a dependency on the Zlib library.
> 
> In that case the libtiff.a file was **NOT** dependent on Zlib, but instead just another listed in the same `target_link_libraries` command. Therefore when CMake generated a Makefile or Ninja ruile file it was free to re-order the listing of those libraries.
> 
> Pretty much the property way to declare a static library dependency can be seen in the [libtiff](https://github.com/o3de/3p-package-source/blob/7901137c1edf74d3f8c7c9fe1081e87cf29a4af2/package-system/tiff/FindTIFF.cmake#L50-L55) cmake today
> 
> #### Proper way to declare dependency: Good
> ```cmake
> # add the CMake standard TIFF::TIFF library.  It is a static library.
> add_library(TIFF::TIFF STATIC IMPORTED GLOBAL)
> set_target_properties(TIFF::TIFF PROPERTIES IMPORTED_LOCATION "${TIFF_LIBRARY}")
> 
> target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB)
> ```
> 
> #### Improper way to declare dependency: Bad
> ```cmake
> # add the CMake standard TIFF::TIFF library.  It is an interface library.
> add_library(TIFF::TIFF INTERFACE IMPORTED GLOBAL)
> 
> target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB "${TIFF_LIBRARY}") # No actual dependency between ZLIB and TIFF and therefore can be reordered
> ```

Oh, I like that idea. These files wouldn't be the right place for CMake education. Would you suggest I update the 3p-package-source repo README.md?

lumberyard-employee-dm:
> > > Please leave some comments so future devs know why this explicit build dependency is here (to resolve issues with linker lib orders)
> > 
> > 
> > Perhaps you should add a link to the 3p-package-source repo with a small blerb explaining about static library link order dependencies. This documentation for the gcc `-l` linker option https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-Libraries explains that the order of linkage is important. Also this stackoverflow answer might help future library maintainers as well https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc/409470#409470.
> > CMake will have static libraries linked in the proper order as long as they are _actually_ marked as STATIC libraries using the [add_library](https://cmake.org/cmake/help/latest/command/add_library.html) function.
> > There was problems early on with O3DE 3rdParty libraries where we we would mark an a 3rdParty library as an `INTERFACE` and then use the `target_link_libraries` command to add both the main file associated with the 3rdParty and then it's dependencies. For example libtiff use to be marked as a INTERFACE library, where it used the `target_link_libraries` function to add the path to the libtiff.a file and a dependency on the Zlib library.
> > In that case the libtiff.a file was **NOT** dependent on Zlib, but instead just another listed in the same `target_link_libraries` command. Therefore when CMake generated a Makefile or Ninja ruile file it was free to re-order the listing of those libraries.
> > Pretty much the property way to declare a static library dependency can be seen in the [libtiff](https://github.com/o3de/3p-package-source/blob/7901137c1edf74d3f8c7c9fe1081e87cf29a4af2/package-system/tiff/FindTIFF.cmake#L50-L55) cmake today
> > #### Proper way to declare dependency: Good
> > ```cmake
> > # add the CMake standard TIFF::TIFF library.  It is a static library.
> > add_library(TIFF::TIFF STATIC IMPORTED GLOBAL)
> > set_target_properties(TIFF::TIFF PROPERTIES IMPORTED_LOCATION "${TIFF_LIBRARY}")
> > 
> > target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > #### Improper way to declare dependency: Bad
> > ```cmake
> > # add the CMake standard TIFF::TIFF library.  It is an interface library.
> > add_library(TIFF::TIFF INTERFACE IMPORTED GLOBAL)
> > 
> > target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB "${TIFF_LIBRARY}") # No actual dependency between ZLIB and TIFF and therefore can be reordered
> > ```
> 
> Oh, I like that idea. These files wouldn't be the right place for CMake education. Would you suggest I update the 3p-package-source repo README.md? Actually, this info seems important event outside of 3p-packages...

It is good educational information for anyone working with static libraries. 
I think it should be added to this repo README.md as it the scenarios where this issue comes up is when a 3rdParty package has linker issues.

When a making a library in the engine or a new gem, a developer would not have problems choosing the correct library type since they are building the code themselves.
So when someone uses an `ly_add_target` function, a developer would not specify an `INTERFACE` target for a STATIC library as the it would need to built. The only way to use the STATIC library in that case would be to build it from source and that requires the target to be specified as STATIC.

This problem comes up when using a pre-built STATIC library, which is specified passing `IMPORTED` to the CMake `add_library` command.

As 3rdParty libraries(and prebuilt Gems) are already built libraries they pass in the `IMPORTED` target to indicate they aren't building from source files, but instead referencing an existing copy of a `lib<name>.a` file on disk.

#### The problem space
The problem is that when we ported 3rdParty libraries to use CMake we didn't use the appropriate library target types for all the actual libraries files.
* `add_library(<libname> STATIC IMPORTED)` for a static library
* `add_library(<libname> SHARED IMPORTED)` for a shared library
* `add_library(<libname> MODULE IMPORTED)` for a runtime loadable-ONLY library

Instead we used the approach of declaring an INTERFACE library target using the form of `add_library(<libname> INTERFACE IMPORTED)`. Afterwards we add a **dependency** for the actual library file (such as `libtiff.a`) on the that INTERFACE library target.
The danger with doing that is that it _slightly_ works in most cases, but can fails if the has dependency on a another library.
Such as libtiff having a dependency on zlib. 

Pretty much the pattern that we were using for defining 3rdParty library targets in CMake was that we were making the main library file a sibling of it's dependencies and not an actual dependency.
So for a library such as libtiff, the setup was previously
> There is a "libtiff" INTERFACE target that depends on two files libtiff.a and libz.a. There is no relation between libtiff.a and libz.a

What we actually wanted was to specify the dependencies explicitly
> There is a "libtiff" STATIC target where it's IMPORTED library file is "libtiff.a". It depends on "libz.a". Therefore CMake will enforce that "libz.a" is listed after "libtiff.a" on the linker line to make sure the "libtiff.a" symbols can be fulfilled

AMZN-Gene:
> > > > Please leave some comments so future devs know why this explicit build dependency is here (to resolve issues with linker lib orders)
> > > 
> > > 
> > > Perhaps you should add a link to the 3p-package-source repo with a small blerb explaining about static library link order dependencies. This documentation for the gcc `-l` linker option https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-Libraries explains that the order of linkage is important. Also this stackoverflow answer might help future library maintainers as well https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc/409470#409470.
> > > CMake will have static libraries linked in the proper order as long as they are _actually_ marked as STATIC libraries using the [add_library](https://cmake.org/cmake/help/latest/command/add_library.html) function.
> > > There was problems early on with O3DE 3rdParty libraries where we we would mark an a 3rdParty library as an `INTERFACE` and then use the `target_link_libraries` command to add both the main file associated with the 3rdParty and then it's dependencies. For example libtiff use to be marked as a INTERFACE library, where it used the `target_link_libraries` function to add the path to the libtiff.a file and a dependency on the Zlib library.
> > > In that case the libtiff.a file was **NOT** dependent on Zlib, but instead just another listed in the same `target_link_libraries` command. Therefore when CMake generated a Makefile or Ninja ruile file it was free to re-order the listing of those libraries.
> > > Pretty much the property way to declare a static library dependency can be seen in the [libtiff](https://github.com/o3de/3p-package-source/blob/7901137c1edf74d3f8c7c9fe1081e87cf29a4af2/package-system/tiff/FindTIFF.cmake#L50-L55) cmake today
> > > #### Proper way to declare dependency: Good
> > > ```cmake
> > > # add the CMake standard TIFF::TIFF library.  It is a static library.
> > > add_library(TIFF::TIFF STATIC IMPORTED GLOBAL)
> > > set_target_properties(TIFF::TIFF PROPERTIES IMPORTED_LOCATION "${TIFF_LIBRARY}")
> > > 
> > > target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > #### Improper way to declare dependency: Bad
> > > ```cmake
> > > # add the CMake standard TIFF::TIFF library.  It is an interface library.
> > > add_library(TIFF::TIFF INTERFACE IMPORTED GLOBAL)
> > > 
> > > target_link_libraries(TIFF::TIFF INTERFACE ZLIB::ZLIB "${TIFF_LIBRARY}") # No actual dependency between ZLIB and TIFF and therefore can be reordered
> > > ```
> > 
> > 
> > Oh, I like that idea. These files wouldn't be the right place for CMake education. Would you suggest I update the 3p-package-source repo README.md? Actually, this info seems important event outside of 3p-packages...
> 
> It is good educational information for anyone working with static libraries. I think it should be added to this repo README.md as it the scenarios where this issue comes up is when a 3rdParty package has linker issues.
> 
> When a making a library in the engine or a new gem, a developer would not have problems choosing the correct library type since they are building the code themselves. So when someone uses an `ly_add_target` function, a developer would not specify an `INTERFACE` target for a STATIC library as the it would need to built. The only way to use the STATIC library in that case would be to build it from source and that requires the target to be specified as STATIC.
> 
> This problem comes up when using a pre-built STATIC library, which is specified passing `IMPORTED` to the CMake `add_library` command.
> 
> As 3rdParty libraries(and prebuilt Gems) are already built libraries they pass in the `IMPORTED` target to indicate they aren't building from source files, but instead referencing an existing copy of a `lib<name>.a` file on disk.
> 
> #### The problem space
> The problem is that when we ported 3rdParty libraries to use CMake we didn't use the appropriate library target types for all the actual libraries files.
> 
> * `add_library(<libname> STATIC IMPORTED)` for a static library
> * `add_library(<libname> SHARED IMPORTED)` for a shared library
> * `add_library(<libname> MODULE IMPORTED)` for a runtime loadable-ONLY library
> 
> Instead we used the approach of declaring an INTERFACE library target using the form of `add_library(<libname> INTERFACE IMPORTED)`. Afterwards we add a **dependency** for the actual library file (such as `libtiff.a`) on the that INTERFACE library target. The danger with doing that is that it _slightly_ works in most cases, but can fails if the has dependency on a another library. Such as libtiff having a dependency on zlib.
> 
> Pretty much the pattern that we were using for defining 3rdParty library targets in CMake was that we were making the main library file a sibling of it's dependencies and not an actual dependency. So for a library such as libtiff, the setup was previously
> 
> > There is a "libtiff" INTERFACE target that depends on two files libtiff.a and libz.a. There is no relation between libtiff.a and libz.a
> 
> What we actually wanted was to specify the dependencies explicitly
> 
> > There is a "libtiff" STATIC target where it's IMPORTED library file is "libtiff.a". It depends on "libz.a". Therefore CMake will enforce that "libz.a" is listed after "libtiff.a" on the linker line to make sure the "libtiff.a" symbols can be fulfilled

Awesome explaination!

Are we stuck in this boat? Adding BUILD_DEP in this PR was because _libaws-cpp-sdk-lambda.a_ (not our library) need to be aware of _aws-cpp-sdk-core_ (also not our library). Our games never directly reference _libaws-cpp-sdk-lambda.a_ though, it always references something we created (3rdParty::AWSNativeSDK::Lamdba).

Can we update that ly_declare_aws_library helper method to what you're suggesting? 

Would that look something like...
```
add_library(3rdParty::AWSNativeSDK::Core STATIC IMPORTED)
set_target_properties(3rdParty::AWSNativeSDK::Core PROPERTIES
    IMPORTED_LOCATION <path-to aws-cpp-sdk-core.a>
)

add_library(3rdParty::AWSNativeSDK::Lamdba STATIC IMPORTED)
set_target_properties(3rdParty::AWSNativeSDK::Lamdba PROPERTIES
    IMPORTED_LOCATION <path-to aws-cpp-sdk-core.a>
    INTERFACE_LINK_LIBRARIES 3rdParty::AWSNativeSDK::Core
)

```

AMZN-Gene:
Will push dependency readme.cd update in a new PR.
https://github.com/o3de/3p-package-source/issues/236

PR #239, opened 2023-12-27
Update Python to 3.10.13

spham-amzn:
Updates to support latest security patches for Python 3.10. Manually fixed broken dependency download links for tcl/tk and ncurses.


package-system/python/darwin_x64/make-python.sh
```diff
@@ -1,4 +1,4 @@
-#
+m#
```
AMZN-alexpete, line 1:
```suggestion
#
```

package-system/python/win_x64/build_python.bat
```diff
@@ -29,16 +29,21 @@ set vswhere_location=%ProgramFiles(x86)%\Microsoft Visual Studio\Installer
 echo adding %vswhere_location% to PATH
 set PATH=%vswhere_location%;%PATH%
 
-for /f "tokens=*" %%i in ('vswhere -version [15.0^,16.0^) -property installationPath') do set VS2017_LOCATION=%%i
+for /f "tokens=*" %%i in ('vswhere -version [16.0^,17.0^) -property installationPath') do set VS2017_LOCATION=%%i
```
AMZN-alexpete, line 32: VS2017??

spham-amzn: The update is to use at least 2019. I just forgot to update the variable name

PR #240, opened 2024-01-24
PhysX Split support for PhysX4

spham-amzn:
Updates to the PhysX 3rd Party  script to support the [PhysX 4/5 Split](https://github.com/o3de/sig-simulation/issues/85).

* Support to rename `FindPhysX.cmake` -> `FindPhysX4.cmake`
* Update the `FindPhysX4.cmake` to only use static libraries instead of the shared equivalent. This is necessary to prevent collision of the same shared libraries from both versions that have the same name. This will also reduce the 3rd Party package size.
* Update the PhysX 3P package to `rev8`


spham-amzn:
Unable to build through GHI due to current build scripts for PhysX not related to these changes.

AMZN-alexpete (approved):

PR #241, opened 2024-01-24
Physx5 updatePhysX Split support for PhysX5

spham-amzn:
Updates to the PhysX 3rd Party script to support the https://github.com/o3de/sig-simulation/issues/85.

* Support to rename FindPhysX.cmake -> FindPhysX5.cmake
* Update the FindPhysX5.cmake to only use static libraries instead of the shared equivalent. This is necessary to prevent collision of the same shared libraries from both versions that have the same name. This will also reduce the 3rd Party package size.
* Fix a macos specific compiile error when using the current macos sdk
* Update the PhysX5 3P package to rev4


spham-amzn:
Unable to build through GHI due to current build scripts for PhysX not related to these changes.

package-system/PhysX5/FindPhysX5.cmake
```diff
@@ -0,0 +1,116 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+# 
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+set(MY_NAME "PhysX5")
```
AMZN-alexpete, line 9:
```suggestion
set(PACKAGE_NAME "PhysX5")
```

package-system/PhysX5/build_package_image.py
```diff
@@ -124,35 +131,55 @@ def cleanUpLibs(self, buildAsStaticLibs):
...
+            if not buildAsStaticLibs:
+
+                for extra_static_lib_filename in ['LowLevel_static_64',
+                                                  'LowLevelAABB_static_64',
+                                                  'LowLevelDynamics_static_64',
+                                                  'PhysXTask_static_64',
+                                                  'SceneQuery_static_64',
+                                                  'SimulationController_static_64']:
+
+                    for extra_static_lib_ext in ['.lib', '.pdb']:
```
AMZN-alexpete, line 147: Are we copying the .pdbs? Do we want to do that? Are they really big? Maybe we could output those to a separate place where they can be zipped and uploaded separate?

AMZN-alexpete (approved):
Just some minor suggestions

package-system/PhysX5/FindPhysX5.cmake
spham-amzn: I'm reluctant to change that pattern now. I think in every 3P package's Find*.cmake it uses the `MY_NAME` pattern.

package-system/PhysX5/build_package_image.py
spham-amzn: This is a common scenario with a lot of these packages. Lets hold off on that until we come up with a general strategy for separate distribution of debug symbols for 3P packages and O3DE

PR #244, opened 2024-01-30
Update Spirv-cross package in order to address some cross compilation bugs for Mac and iOS

moudgils:
The current version of SPIRV-Cross is throwing exceptions when trying to compile spirv into metalSL which is addressed by the latest version.


package-system/SPIRVCross/build_config.json
```diff
@@ -1,29 +1,16 @@
...
-               "x64",
-               "-T",
-               "host=x64",
-               "-DSPIRV_CROSS_CLI=ON",
-               "-DSPIRV_CROSS_SHARED=OFF",
-               "-DCMAKE_INSTALL_LIBDIR=\"lib/Debug\"",
-               "-DCMAKE_INSTALL_BINDIR=\"bin/Debug\"",
-               "-DCMAKE_BUILD_TYPE=Debug"
-            ],
             "cmake_generate_args_release":[
```
AMZN-alexpete, line 14: Minor recommendation to remove the Generator for Windows
```
               "-G",
               "\"Visual Studio 16 2019\"", 
```
CMake will use the most recent version of Visual Studio on the build machine which is what we did on our Jenkins machines so we don't have to update these files every time we update the build nodes with a new version of Visual Studio.

Note that for platforms like Linux you do want to keep the `G Ninja` because it might default to GCC, and I think you need to leave `Xcode` for Mac too.

AMZN-alexpete (approved):
just a minor suggestion to remove the generator arg for Windows to always use the latest version of VS on the build node.

package-system/SPIRVCross/build_config.json
moudgils: Yep. I had to make this change as the node didn't have VS2019

package-system/SPIRVCross/build_config.json
```diff
@@ -1,32 +1,17 @@
...
-               "-A",
-               "x64",
-               "-T",
-               "host=x64",
-               "-DSPIRV_CROSS_CLI=ON",
-               "-DSPIRV_CROSS_SHARED=OFF",
-               "-DCMAKE_INSTALL_LIBDIR=\"lib/Debug\"",
-               "-DCMAKE_INSTALL_BINDIR=\"bin/Debug\"",
-               "-DCMAKE_BUILD_TYPE=Debug"
-            ],
```
spham-amzn, line 26: Why was the entire `cmake_generate_args_debug` removed? Its causing the `pull_and_build_from_git` script to fail. You should only need to remove the `-G`, `-A`, `-T`, and `host=x64` arguments

package-system/SPIRVCross/build_config.json
```diff
@@ -42,16 +27,6 @@
...
-            "cmake_generate_args_debug":[
-               "-G",
-               "Xcode",
-               "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.13",
-               "-DSPIRV_CROSS_CLI=ON",
-               "-DSPIRV_CROSS_SHARED=OFF",
-               "-DCMAKE_INSTALL_LIBDIR=\"lib/Debug\"",
-               "-DCMAKE_INSTALL_BINDIR=\"bin/Debug\"",
-               "-DCMAKE_BUILD_TYPE=Debug"
-            ],
```
spham-amzn, line 54: Was this needed to be removed? This is causing the script to fail, it needs at least the generator `-G` argument

package-system/SPIRVCross/build_config.json
```diff
@@ -67,15 +42,6 @@
...
             "cmake_find_template":"FindSPIRVCross.cmake.Linux",
-            "cmake_generate_args_debug":[
-               "-G",
-               "Ninja",
-               "-DSPIRV_CROSS_CLI=ON",
-               "-DSPIRV_CROSS_SHARED=OFF",
-               "-DCMAKE_INSTALL_LIBDIR=\"lib/Debug\"",
-               "-DCMAKE_INSTALL_BINDIR=\"bin/Debug\"",
-               "-DCMAKE_BUILD_TYPE=Debug"
-            ],
```
spham-amzn, line 78: Same comment here

package-system/SPIRVCross/build_config.json
amzn-changml: If the intention is to only build on the release target, then a `--build_configs release` argument should be added to each `package_build_list_host_*.json`command. 

`pull_and_build_from_git.py` defaults to the release and debug target if no argument is given: https://github.com/o3de/3p-package-source/blob/main/Scripts/extras/pull_and_build_from_git.py#L279C11-L279C103

moudgils: I brought the debug configs back to allow the packages to be built.

package-system/SPIRVCross/build_config.json
moudgils: I have re-added it back as a debug config is needed.

package-system/SPIRVCross/build_config.json
moudgils: Same

moudgils:
This is the autogenerated PR for runtime - https://github.com/o3de/o3de/pull/17446. Will merge both PRs together

spham-amzn (approved):

PR #266, opened 2024-11-21
Add support for a wayland variant of Qt for Linux

spham-amzn:
- Remove unnecessary wrapper scripts build-linux-aarch64.sh and build-linux-x86.sh
- Add special build_config_wayland.json and Dockerfile.wayland to support the wayland variant
- Update docker_build_qt_linux.sh to support the environment option 'ENABLE_QT_WAYLAND' which is set in the Dockerfile
- Update docker_build_qt_linux.sh to support building the additional qtwayland package from source


DrogonMar:
We could try enabling both xcb and wayland QPA plugins

DrogonMar:
Why do we disable EGL for QT [here](https://github.com/o3de/3p-package-source/blob/f7e8554e2681b35599d96c87a836888f3f5ea8f6/package-system/Qt/docker_build_qt_linux.sh#L72C22-L72C25)?

DrogonMar:
Why do we disable EGL in QT [here](https://github.com/o3de/3p-package-source/blob/f7e8554e2681b35599d96c87a836888f3f5ea8f6/package-system/Qt/docker_build_qt_linux.sh#L72C22-L72C25)?

*Sorry for duplicate message, thought I could put a message on the code instead of a comment here*

DrogonMar:
I'm not sure on how building QT works so let me know if its just because it allows it to work or not but why are you enabling wayland-server?
A Wayland server is like Windows DWM (I dont know if this is correct but Window's component that handles drawing windows to the display), KDE or Gnome.
O3DE only needs to be a Wayland client.

DrogonMar:
> I don't remember, my guess is that we didn't want to build the [eglfs](https://doc.qt.io/qt-6/embedded-linux.html#embedded-eglfs) plugin?

Hmm, could we explicitly disable that and just allow normal EGL?
I would like to have the QT interface GPU accelerated.

DrogonMar:
Awesome, [that did work](https://github.com/o3de/3p-package-source/actions/runs/12383558047/job/34566505007#step:12:4713) but we should disable the compositor/server part of it as thats under GPL3 which isn't good and we dont need it, look at the info [here](https://doc.qt.io/qt-5/qtwaylandcompositor-index.html#licenses-and-attributions).

package-system/Qt/docker_build_qt_linux.sh
```diff
@@ -69,7 +69,6 @@ echo Configuring Qt  ...
                  -no-feature-getentropy \
                  -no-feature-statx \
                  -feature-wayland-server \
```
DrogonMar, line 71: We should remove the wayland server feature as its under GPLv3 and we don't need it

DrogonMar:
It's still [enabling the compositor feature](https://github.com/o3de/3p-package-source/actions/runs/12396903156/job/34606093438?pr=266#step:12:4732), maybe `-no-feature-wayland-server` going by this [doc.](https://doc.qt.io/qt-5/configure-options.html#including-or-excluding-features)

spham-amzn:
@amzn-changml Is there a way to get the ARM64 GHA build to run this again?

DrogonMar (commented):
Looks good

amzn-changml:
The arm64 build is fixed, just need to merge from latest

package-system/Qt/Dockerfile
```diff
@@ -8,12 +8,39 @@
...
+ARG INPUT_DOCKER_BUILD_SCRIPT
+ARG INPUT_BUILD_FOLDER
+ARG INPUT_DEPENDENT_PACKAGE_FOLDERS
+
+ENV WORKSPACE=/data/workspace
+ENV DOCKER_BUILD_PATH=$WORKSPACE/$INPUT_BUILD_FOLDER
+ENV DOWNLOADED_PACKAGE_FOLDERS=$INPUT_DEPENDENT_PACKAGE_FOLDERS
+
+WORKDIR $WORKSPACE
+
```
byrcolin, line 37: Is this this correct? Setting and then setting again right after?

byrcolin (approved):
Looks ok, just unsure about the WORKDIR $WORKSPACE and then WORKDIR /data/workspace right after it. If thats correct or makes no difference then merge it.

PR #269, opened 2024-12-13
Forcing New Assimp Artifacts

AMZN-Gene:
Bumping version number so that github actions produces new artifacts; the last assimp PR didn't contain a version bump because the package wasn't actually updating in o3de, but a bump is required for github actions to pick up on the change.


amzn-changml (approved):

spham-amzn:
@amzn-changml Does this need to be bumped again to rebuild the 3P packages?

guillaume-haerinck:
Might be overseeded by https://github.com/o3de/o3de/pull/19365

PR #281, opened 2025-06-07
Update crashpad library to latest to allow linux build

guillaume-haerinck:
Related to https://github.com/o3de/o3de/issues/6000. 

# What ?

The [crashpad library](https://chromium.googlesource.com/crashpad/crashpad) is used to generate dump files for your process when said process is about to crash. The current version of the library only builds on Windows. With this update to the latest version [available via vcpgk](https://vcpkg.link/ports/crashpad) it can build on android, linux and ios and mac as well.

# Changes

This PR keep the windows build, but also introduce the linux build target. Other platform are to be introduced in future PR.

## What does the patchfile contains ?

I kept the library modification originally introduced by o3de, for crashpad, it contains hooks to :

- Attach new content before sending the report to a server (`AddAttachments` and `UpdateHttpTransport`)
- Destroy the report if we want to (`CheckConfirmation`)

I also introduced a new hook on client to be able to react to a crash from the crashing process. As we are not currently sending over the dump file, I will use it to create a log file with the callstack that user can submit to github.

There are also modifications on the [vcpkg portfile](https://github.com/microsoft/vcpkg/blob/5c41ff21f6feda702a7d5a1afc765f88d25783c5/ports/crashpad/portfile.cmake) to modify the built target list. Instead of building the crash handler executable as an external application, we build its dependencies as a set of static libraries and do the building of the handler executable on o3de side.

## Other changes

As, in order to apply the patchfile, its line-endings have to be LF instead of CRLF, I've made that explicit via a gitattribute file to prevent issues (I took me a bit to understand that the patchfile wasn't applied because of line endings).

# How to test

> In order to debug the build, you might wanna change the `tempdir` variable in `package-system/Crashpad/build_package_image.py` to an actual existing repository

## On windows

With this repository and the [3p-package-scripts](https://github.com/o3de/3p-package-scripts) repository. From the `o3de_package_scripts` folder run :

```bash
python .\build_package.py  --search_path ..\..\3p-package-source Crashpad-0.8.0-rev2-windows
```

Copy the generated file `3p-package-source\packages\Crashpad-0.8.0-rev2-windows.tar.xz` to your third party location. By default `USER\.o3de\3rdParty\downloaded_packages`

Then with a source build of the engine. Modify :

- `o3de\cmake\3rdParty\Platform\Windows\BuiltInPackages_windows.cmake` : Update crashpad name and hash with what was provided by the library build
- `o3de\cmake\Platform\Windows\PAL_windows.cmake` : Update `PAL_TRAIT_BUILD_EXTERNAL_CRASH_HANDLER_SUPPORTED` to true

Rebuild, launch the editor, and in the console window, type "crash" and press enter. The error handler dialog should open.

## On linux

The setup is similar to the windows one, however I had to do **`sudo apt install python3-venv libcurl4-openssl-dev` for the build to work**. I don't know if this can be worked around via vcpkg.

Then to actually test the package on the linux engine, multiple cmake files are missing and needs to be added on the CrashHandler library on o3de side. Once everything was setup, **I did encounter link issues with crashpad that I wasn't able to resolve yet** (the crashpad minidump library not finding symbols in the crashpad snapshot library).

I still decided to include the linux build inclusion to reduce the future efforts needed to make crashpad work on o3de linux


nick-l-o3de (approved):
LGTM - crashpad is definitely one of those libraries I'd like to eventually convert to an inline built one if possible, but for now... we keep it this way I guess :)

PR #292, opened 2025-09-01
Build DXC with the VS2019 toolset

msat-huawei:
There are some problems with the Jenkings CI where DXC complains about missing dlls.
O3DE currently supports building with VS2019 so the VS2022 runtime should probably not be required.


msat-huawei:
Unfortunately this does not seem to build on Windows as the VS2019 toolkit does have the ATL libraries installed.
According to this https://github.com/actions/runner-images/issues/12287 the libraries should now be installed, but apparently they aren't.

msat-huawei:
I managed to build it on Windows with VS2019 by manually installing ATL as described in https://github.com/actions/runner-images/issues/12486
This does take ~10min more per build, but I currently don't see a way around that.

amzn-changml (approved):

PR #297, opened 2026-01-07
Mac ARM64 3P support : Initial setup for Mac ARM64

spham-amzn:
* Create new package_build_list* file for darwin-arm64
* Update Mac Toolchain to deployment target of 15
* Update build-package.yaml to use os 'macos-15'

### DIFF
--- a/package_build_list_host_darwin-arm64.json
+++ b/package_build_list_host_darwin-arm64.json
@@ -1,0 +1,10 @@
+{
+    "comment" : "This is the list that applies to darwin arm64 (Mac Silicon) hosts",
+    "comment2" : "build_from_source is package name --> build script to call with params",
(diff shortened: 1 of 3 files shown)


package_build_list_host_darwin-arm64.json
```diff
@@ -0,0 +1,10 @@
+{
+    "comment" : "This is the list that applies to darwin arm64 hosts",
```
amzn-changml: It seems the 3p build script GHA is doing weird things with the comment strings. Should try to fix that first before we merge this in, or remove this comment block for now?

amzn-changml:
The commit message for this PR is called `Mac ARM64 3P support : Add definition for zlib-1.3.1-rev1-mac-arm64`. Should we rename that to be more accurate?

amzn-changml (approved):

PR #329, opened 2026-01-13
Remove v-hacd from 3P packaging system

spham-amzn:
* v-hacd-2.3-1a49edf has been deprecated and updated to v4.1.0 (ref https://github.com/o3de/o3de/pull/18718) (Thank you @[rossbridger](https://github.com/rossbridger)) 
* Removed all references to v-hacd


amzn-changml:
Note that this package error is due to the package being removed and the build script not having anything to do.  This is non-blocking, however, since there's nothing to promote in this case. You're free to merge once this PR is approved.

We'll fix this case in the next optimization pass for the 3P script.

amzn-changml (approved):

PR #361, opened 2026-02-14
Fix astc-encoder Findastc-encoder.cmake error for Mac ARM64

spham-amzn:
The Findastc-encoder.cmake for Mac ARM64 was using the combined template source that was loading the x86 version of the static library, not the arm64 (neon) version.


amzn-changml (approved):
Promotion run here https://github.com/o3de/3p-package-source/actions/runs/22025924476

O3DE PR here https://github.com/o3de/o3de/pull/19569

PR #369, opened 2026-03-29
Build Pyside6

guillaume-haerinck:
Related to https://github.com/o3de/o3de/pull/19567, allow to build pyside6, which allow to use our Qt6 version with python tools.

## Where was it tested ?

- Working on windows
- Pending testing on linux as build fails locally (opened this PR to try the build on the CI side)


spham-amzn (approved):
Holding off on the Mac for now, since it was never supported.

PR #373, opened 2026-05-15
Qt6 add patch file for recursion check

spham-amzn:
- Taking over PR: https://github.com/o3de/3p-package-source/pull/368 from https://github.com/guillaume-haerinck
- Updating CLANG_VER from 12 to 15


guillaume-haerinck (approved):

.github/workflows/build-package.yaml
```diff
@@ -280,7 +280,7 @@ jobs:
     - name: Install clang/gcc
       if: runner.os == 'Linux'
       env:
-        CLANG_VER: 12
+        CLANG_VER: 15
```
amzn-changml, line 283: Just curious, is this the minimum version we want to target or just for QT going forward? Just wondering why we wouldn't want to go up to 16 or 18? No issues with this version otherwise for the lowest common denominator

amzn-changml (commented):
Will start the promotion process and o3de PR AR first before approval

.github/workflows/build-package.yaml
spham-amzn: There was related to some discussions in the Qt6 channel related to Linux. It is set to 15 since that was the latest version supported by Ubuntu 22.04 through apt.

PR #381, opened 2026-06-06
RPATH fix for pyside6 on Linux

spham-amzn:
Apply a linux pre-package RPATH fix to 

 - bin/shiboken6
 - lib/libpyside6.abi3.so.6.10
 - lib/libpyside6qml.abi3.so.6.10
 - lib/libshiboken6.abi3.so.6.10

Fixes https://github.com/o3de/3p-package-source/issues/378


package-system/pyside6/build_config.json
```diff
@@ -29,6 +29,7 @@
         },
         "Linux": {
             "Linux": {
+                "package_version": "6.10.2-py3.10-rev2",
                 "depends_on_packages" :[
                     ["python-3.10.13-rev2-linux", "a7832f9170a3ac93fbe678e9b3d99a977daa03bb667d25885967e8b4977b86f8", ""],
                     ["qt-6.10.2-rev5-linux", "1771f261060e33b90312aee05647d17d5f24c59d61834e09b7e98f3065058d31", ""]
```
guillaume-haerinck: We should make it depend on qt-6.10.2-rev6-linux from https://github.com/o3de/3p-package-source/pull/380/changes out of safety

nickschuetz (commented):
Nice, this was fast! I pulled the actual package your PR build produced (run 27052719627) and checked it over. The four files you patched all come out exactly right (`$ORIGIN` on bin/shiboken6, `$ORIGIN:$ORIGIN/../shiboken6` on the libs), and the `.so` symlinks survive packaging fine.

While I was in there I scanned the rest of the ELFs and there are still a few stragglers with the old runpath: the `Qt*.abi3.so` modules under `lib/python3.10/site-packages/PySide6/` (19 of them), plus one neither of us had spotted before, a second copy of shiboken6 under `shiboken6_generator/`. Worth saying: for the engine itself your four are already enough (only bin/shiboken6 and libpyside6.abi3.so.6.10 actually ship into the engine, and those were the only two our packaging checks ever complained about), so the rest is just nice-to-have package hygiene.

If you want to catch them all without playing whack-a-mole, something like this is what we do on our side and it covers anything that shows up in the future too:

```bash
find "$PACKAGE_BASE" -type f \( -name '*.so*' -o -perm -111 \) | while read -r f; do
    rp=$(patchelf --print-rpath "$f" 2>/dev/null) || continue
    [ -z "$rp" ] && continue
    clean=$(printf '%s' "$rp" | tr ':' '\n' | grep '^\$ORIGIN' | paste -sd:)
    [ "$rp" != "$clean" ] && patchelf --set-rpath "$clean" "$f"
done
```

(it just drops the absolute entries and keeps each file's own `$ORIGIN` ones). Totally optional thought: now that #380 is merged, the rev6 qt package is one promote approval away, so the qt pin in `depends_on_packages` could point at `qt-6.10.2-rev6-linux` and save a touch later.

Either way, happy to re-check the rebuilt artifact whenever, it takes me a couple minutes. Also happy to approve as-is if you'd rather land the engine-path fix now and handle the stragglers separately, your call, just say the word.

package-system/pyside6/build_config.json
spham-amzn: Sure, once it gets published I can update

guillaume-haerinck (approved):

PR #384, opened 2026-06-15
Fix pyside6 shiboken issues

spham-amzn:
- Fix shiboken flags to use c++20 to match o3de and fix errors
- Fix the RPATH settings for shiboken6 to work with using $QTDIR/bin as the current working directory (set to a relative ./../lib) for linux and mac to resolve libQtCore.so.6
- Add missing copy of the share folder into the package (needed for shiboken6)
- Bump the revision to 4
- Tagged the windows github runner image to windows-2022 to lock in visual studio 2022


package-system/pyside6/build_config.json
```diff
@@ -43,7 +43,6 @@
         },
         "Linux": {
             "Linux": {
-                "package_version": "6.10.2-py3.10-rev3",
                 "depends_on_packages" :[
                     ["python-3.10.13-rev2-linux", "a7832f9170a3ac93fbe678e9b3d99a977daa03bb667d25885967e8b4977b86f8", ""],
                     ["qt-6.10.2-rev6-linux", "34dcdc366e701f6cea1b6d06013ed6aa98521a6d663d244612ebcad3eae7ce34", ""]
```
guillaume-haerinck, line 48: We are at rev7 now on linux if I believe the root file

guillaume-haerinck (approved):

package-system/pyside6/build_config.json
spham-amzn: rev6 and rev7 won't have any ABI breaking differences, so it shouldn't matter. It only had updates to which Qt modules it exported.

.github/workflows/build-package.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
                 PACKAGE_OS="ubuntu-22.04"
                 echo "OS Image selected for $PACKAGE: $PACKAGE_OS"
               elif [[ $PLATFORM =~ "windows" ]]; then
-                PACKAGE_OS="windows-latest"
+                PACKAGE_OS="windows-2022"
```
amzn-changml, line 75: This is fine, and looks like latest has broken VS paths. Could you do me a favor and call this out in this PR, so we know when this was set?

amzn-changml (approved):

PR #385, opened 2026-06-21
Additional Pyside6/Qt6 fixes

spham-amzn:
## Pyside6/Shiboken Fixes

* Add compiler-path to Shiboken6 arguments to fix linux build error
* Fix extension for generated files from shiboken: pyd is windows only, .so is linux/mac

## Qt6 Fixes
* Adjust lrelease RPATH for special case where lrelease is copied to the O3DE bin folder
* Suppress warning about using a private Qt header
* Remove deprecated qt5.15 from package reference for linux


package-system/Qt/copy_platform_cmakes.py
```diff
@@ -49,3 +49,14 @@
         sys.exit(1)
     print(f"Copying {file} => {package_qt_root}")
     shutil.copy2(file, package_qt_root)
+
+
+if platform_system == "linux":
+    # On Linux, we need to fix the rpath for lrelease
+    lrelease_path = pathlib.Path(package_root) / "qt" / "bin" / "lrelease"
```
guillaume-haerinck: this rpath change is not required since its done via https://github.com/o3de/o3de/pull/19852 (and I think we'll drop that in a near future)

package-system/Qt/Platform/Linux/Qt_linux.cmake
```diff
@@ -8,6 +8,15 @@
...
+#  This project is using headers of the GuiPrivate module and will therefore
+#  be tied to this specific Qt module build version.  Running this project
+#  against other versions of the Qt modules may crash at any arbitrary point.
+#  This is not a bug, but a result of using Qt internals.  You have been
+#  warned!
+set(QT_NO_PRIVATE_MODULE_WARNING ON)
+
 list(APPEND QT6_COMPONENTS
     DBus
     GuiPrivate
```
guillaume-haerinck, line 22: To re-evaluate once wayland support is stabilized as we might not need to depend on private QT

guillaume-haerinck (approved):

package-system/Qt/copy_platform_cmakes.py
spham-amzn: I agree, I think https://github.com/o3de/o3de/pull/19852 is the right place for the rpath change. I'll remove this

nickschuetz (approved):

PR #387, opened 2026-07-20
Bump OpenSSL 1.1.1 -> 3.6.3

jhanca-robotecai:
## Bump OpenSSL to 3.6.3 across all platforms

Unifies the O3DE OpenSSL 3rdParty package to 3.6.3, replacing the 1.1.1.

### Changes
-  Linux, Linux-aarch64, and Mac/Mac-arm64: source builds now target `openssl-3.6.3`.
-  vcpkg path (Mac-Intel/iOS/Windows/Android): bumped the vcpkg pin to a commit shipping OpenSSL 3.6.3 natively; dropped the two legacy 1.1.1-era patches (`set_openssl_port_to_1_1_1_x.patch`, `enable-stdio-on-iOS.patch`); updated package metadata (OpenSSL changed to `Apache-2.0`).
- fixed two stale prerequisite checks in testing script: the `qemu` package no longer exists on modern Ubuntu, and the aarch64 binfmt check now also recognizes kernel-level `binfmt_misc` registration

### Testing
- Linux x86_64: built and tested end-to-end via the real Docker pipeline
- vcpkg path: validated via a local Linux-triplet dry run of the new vcpkg pin
- Linux-aarch64, Mac/Windows/Android/iOS: left to CI


nick-l-o3de (approved):
Looks fine by me!  Note that last I checked, it was used by Python, and Qt and a few other libraries, notably

* AWSGameLift stuff - and other AWS stuff - we don't maintain this anymore, so its not relevant. We should, in fact, consider eliminating any related folders.  They'll be there in history, but we can close it up for now.
* Qt - note how the Qt build config.json refers to OpenSSL-1.1.1t-rev1-linux and windows.  We may have to rebuild Qt to ensure they're all on the same platform here.
* Python - Python has crypto built into it because it is everything AND the kitchen sink.  We may have to ship a new python, and thus a new Qt For Python (pyside6) lib too.
* O3DE uses it as far as I know in the networking platform.

PR #388, opened 2026-07-21
Modernize Android 3p build scripts for current NDK/CMake

byrcolin:
## What does this PR do?

Fixes the Android 3p package build scripts so they work with current NDK and CMake versions. No functional change to package outputs.

- **Toolchain**: opt out of the NDK's legacy toolchain wrapper (`ANDROID_USE_LEGACY_TOOLCHAIN_FILE OFF`), which is broken with CMake 4.x
- **mikkelsen**: remove hardcoded NDK r21d path (uses the `LY_NDK_DIR` env var, which the toolchain already supports), drop the API level pin (toolchain enforces and defaults to 24), remove obsolete arm32 flags (`ANDROID_ARM_MODE`/`ANDROID_ARM_NEON` modern NDKs hard-error on disabling NEON, and both are meaningless for arm64-v8a)
- **zlib**: pass `CMAKE_POLICY_VERSION_MINIMUM=3.5` so zlib's ancient `cmake_minimum_required(VERSION 2.4.4)` configures under CMake 4.x (must be on the command line the top-level version check runs before the toolchain file is read)
- **.gitignore**: exclude `build/` folders

## How was this tested?

Built with NDK r27.3.13750724 / CMake 4.1 on Windows:
- mikkelsen: debug + release via Ninja Multi-Config, both installed correctly
- zlib: `build_zlib_android.cmd` links `libz.a` cleanly

## Related

Prerequisite for the Android 16KB page alignment work: RFC o3de/o3de#19936

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ TestResults/**
 .pytest_cache
 packages/
 **/temp/**
+**/build/**
 qt_compile_log.txt
 .remote_cache.json
 .venv/
--- a/Scripts/cmake/Platform/Android/Toolchain_android.cmake
+++ b/Scripts/cmake/Platform/Android/Toolchain_android.cmake
@@ -58,6 +58,9 @@ set(CMAKE_CXX_FLAGS "-fPIC")
 # Make a backup of the CMAKE_FIND_ROOT_PATH since it will be altered by the NDK toolchain file and needs to be restored after the input
 set(BACKUP_CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH})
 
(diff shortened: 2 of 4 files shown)


nick-l-o3de (approved):
LGTM.  We should be getting those vars from the YAML for the CI so there's no need to set it. 
```yaml
 # Set NDK folder path in the env vars
        $ndkRoot = "$sdkPath\ndk\$env:ANDROID_NDK_VERSION"
        echo ANDROID_NDK=$ndkRoot >> $env:GITHUB_ENV
        echo ANDROID_NDK_HOME=$ndkRoot >> $env:GITHUB_ENV
        echo ANDROID_NDK_ROOT=$ndkRoot >> $env:GITHUB_ENV
```

PR #391, opened 2026-07-24
Follow-ups to OpenSSL 3.6.3 bump: CPython

jhanca-robotecai:
### Qt: remove unused script parameters (Linux/Linux-aarch64)
The `OPENSSL_FOLDER_NAME` argument in `build-linux.sh` was dead code, never referenced after assignment. Removed the dependency and the unused parameter; Windows's OpenSSL usage is untouched. The same for two other parameters.

### Python: rebuild bundled OpenSSL to 3.6.3 (Linux/Linux-aarch64)
`_ssl.cpython-310-x86_64-linux-gnu.so` and `_hashlib.cpython-310-x86_64-linux-gnu.so` statically embed and export **unhidden** OpenSSL symbols: update the dependency from  1.1.1t to 3.6.3.

Windows and Mac Python are unaffected by this PR, both are stuck on CPython 3.10's own upstream-frozen OpenSSL 1.1.1u/1.1.1w.

### Testing
- Downloaded and inspected the actual published `qt-*-linux`/`qt-*-windows` and `python-*-linux` packages (`ldd`, `nm -D`, `strings`)
- Local build/test not yet run for the Python rebuild: relying on CI


spham-amzn:
@jhanca-robotecai There seems to be a problem deploying the OpenSSL to the dev environment. The build is no longer there, so we may need to do a revision bump to get another build of OpenSSL and then promote it as soon as the jobs finish building.

PR #398, opened 2026-08-07
Upgrades to pull_and_build_from_git.py to support upcoming libraries

nick-l-o3de:
## WHAT DO

* adds `clean_between_configurations` option and docs for it!
  * Justification: some libraries like openimageio cache their own fetchcontent dependencies in a single location during **configure** instead of build, and then don't rebuild them regardless of the configuration requested, causing release to compile against debug dependencies, for example, if debug was configured first.  This would not matter on windows if you could use the release dependencies in debug, or vice versa, but unfortunately, OpenImageiO depends on C++ libraries, not just C API libraries, which then causes iterator mismatches.
* adds `set_env_vars` option, which allows build scripts to set additional environment variables for the build, as they need.
* `additional_src_files` was not documented, but already existed.
   * added docs for it.
   * moved the `additional_src_files` copy operation until **after** all the git checkout and package downloading has happened, so that the additional files can copy/override the git checkout and package downloading (like a patch if necessary).  OpenImageIO needed this because it asks for ZLIB in a very specific and irritating way, thus needing to patch the `FindZLIB.cmake` file.
* Adds a check for if delete folder is about to delete a root folder of the machine, just in case.
* There's already a function to apply a custom env var set, but it was only being applied during test.  Now applies the custom environment variables requested to the configure and build steps, not just the test step, so that it is a uniform environment.
* Adds `CMAKE_PREFIX_PATH` environment variable to any time the custom environment variable is used, and has it include the downloaded packages, so that the build can find them if they are requested via `find_package(CONFIG...)` which will not make use the `CMAKE_MODULE_PATH`.

## HOW TEST

I've been using this upgraded script for a few weeks, to build Imath, OpenEXR, OpenImageIO, libTIFF, YAML-CPP, etc, they all rely on it, so its been copied into each of them.  I  figure that instead of doing that, we should just commit it once to the 3p repo, then they can all stack on top of it.

### DIFF
--- a/Scripts/extras/pull_and_build_from_git.py
+++ b/Scripts/extras/pull_and_build_from_git.py
@@ -92,6 +92,10 @@ The following keys can exist at the root level or the target-platform level:
                                             install tree will be copied to the target package.
                                             This field can exist at the root but also at individual platform target level.
 
+* clean_between_configurations            : (optional), default False.  If True, the build folder will be erased between compiling
+                                            different configurations.
+
+* set_env_vars                            : dictionary of env vars to set { "key" : "value" , ... }
 
 The following keys can only exist at the target platform level as they describe the specifics for that platform.
 
@@ -155,6 +159,18 @@ The following keys can only exist at the target platform level as they describe 
                                                file hash - The hex-string of the fingerprint to validate the download with. If this is left blank, no validation
                                                            will be done, instead it will be calculated on the downloaded package and printed to the console.
                                                hash algorithm - The hash algorithm to use to calculate the file hash.
+
+* additional_src_files                   : individual files to copy from the local folder into
+                                           the temp src folder before build.
+                                           elements in this list can either be a string
+                                           filename,
+                                           or a list of 2 elements, the first being the string
+                                           filename, and the second being a relative path from
+                                           temp.
+                                           E.g:
+                                           [  "extrafile.txt", "anotherfile.txt" ] - copied into temp/src/anotherfile.txt
+                                           [ [ "extrafile.txt", "dest/foo.txt"],   - copied into temp/dest/foo.txt
+                                           [ "anotherfile.txt" ] - copied into temp/src/bar.txt
 
 
 Note about environment variables:
@@ -276,7 +292,7 @@ class PackageInfo(object):
         self.git_commit = _get_value("git_commit", required=False)
         self.cmake_find_template = _get_value("cmake_find_template", required=False)
         self.cmake_find_source = _get_value("cmake_find_source", required=False)
-        self.cmake_find_target = _get_value("cmake_find_target")
+        self.cmake_find_target = _get_value("cmake_find_target", required=False)
         self.cmake_find_template_custom_indent = _get_value("cmake_find_template_custom_indent", default=1)
         self.additional_src_files = _get_value("additional_src_files", required=False)
         self.depends_on_packages = _get_value("depends_on_packages", required=False)
@@ -288,6 +304,8 @@ class PackageInfo(object):
         self.extra_files_to_copy = _get_value("extra_files_to_copy", required=False)
         self.cmake_install_filter = _get_value("cmake_install_filter", required=False, default=[])
         self.custom_toolchain_file = _get_value("custom_toolchain_file", required=False)
+        self.clean_between_configurations = _get_value("clean_between_configurations", required=False, default=False)
+        self.set_env_vars = _get_value("set_env_vars", required=False, default={})
 
         if self.cmake_find_template and self.cmake_find_source:
             raise BuildError("Bad build config file. 'cmake_find_template' and 'cmake_find_source' cannot both be set in the configuration.")
@@ -394,6 +412,9 @@ def delete_folder(folder):
         print(f"Expected a folder, but found a file: {path_folder}")
     if not path_folder.is_dir():
         return
+
+    if os.path.dirname(path_folder) == path_folder: # is it the root folder?
+        raise BuildError(f"Was told to erase folder {path_folder} but that seems like a bad idea.")
 
     if platform.system() == 'Windows':
         call_result = subprocess.run(subp_args(['rmdir', '/Q', '/S', str(path_folder)]),
@@ -626,17 +647,6 @@ class BuildInfo(object):
         else:
             raise BuildError(f"Missing both 'git_url' and 'src_package_url' from the build config")
 
-
-        if self.package_info.additional_src_files:
-            for additional_src in self.package_info.additional_src_files:
-                additional_src_path = self.base_folder / additional_src
-                if not additional_src_path.is_file():
-                    raise BuildError(f"Invalid additional src file: : {additional_src}")
-                additional_tgt_path = self.src_folder / additional_src
-                if additional_tgt_path.is_file():
-                    additional_tgt_path.unlink()
-                shutil.copy2(str(additional_src_path), str(additional_tgt_path))
-
         # Check/Validate the license file from the package, and copy over to install path
         if self.package_info.package_license_file:
             package_license_src = self.src_folder / self.package_info.package_license_file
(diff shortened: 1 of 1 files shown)


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -704,9 +714,26 @@ def sync_source(self):
                 extracted_package_path = extract_package(src_package_file=downloaded_package_file,
                                                          target_folder=self.base_temp_folder)
 
+        # finally, copy the additioanl src files if any, so they can overwrite anything above
```
EnterTheArcane: Could we move the license validation/copy below this overlay step? I tested the patched script with a license supplied through additional_src_files, and it exits with Invalid/missing license file before reaching this code. This also affects the existing SQLite, Lua, squish-ccr, and zlib configs, which use additional_src_files to supply their package_license_file. As written, those package builds will fail.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -395,6 +413,9 @@ def delete_folder(folder):
     if not path_folder.is_dir():
         return
 
+    if os.path.dirname(path_folder) == path_folder: # is it the root folder?
```
EnterTheArcane: I think this guard compares different types. path_folder is a Path, while os.path.dirname(path_folder) returns a string, so the condition is false even for /. I intercepted subprocess.run and confirmed that delete_folder("/") continued on to construct rm -rf /. Could use `path_folder.parent == path_folder` instead?

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -276,7 +292,7 @@ def _get_value(value_key, required=True, default=None):
         self.git_commit = _get_value("git_commit", required=False)
         self.cmake_find_template = _get_value("cmake_find_template", required=False)
         self.cmake_find_source = _get_value("cmake_find_source", required=False)
-        self.cmake_find_target = _get_value("cmake_find_target")
+        self.cmake_find_target = _get_value("cmake_find_target", required=False)
```
EnterTheArcane: Is making this optional intentional? I tested a config without cmake_find_target, PackageInfo accepts it, but generate_cmake later evaluates package_install_root/None and raises a TypeError. Unless a no-target flow is being added, this should probably remain required so the config fails early with a useful error.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -704,9 +714,26 @@ def sync_source(self):
...
 
+        # finally, copy the additioanl src files if any, so they can overwrite anything above
+        if self.package_info.additional_src_files:
+            for additional_src in self.package_info.additional_src_files:
+                additional_src_src = ""
+                additional_src_tgt = ""
 
+                if isinstance(additional_src, list):
+                    additional_src_src = additional_src[0]
+                    additional_src_tgt = self.base_temp_folder / additional_src[1]
```
EnterTheArcane: Could we validate and prepare the destination here? I confirmed that a `../outside-temp.txt` destination writes outside base_temp_folder, while an absolute destination replaces the base path entirely. I also tested the documented nested-destination form, and `dest/subfolder/file.txt` raises FileNotFoundError when its parent does not already exist. It might be a good idea to require exactly two strings, resolve the target, check that it remains under base_temp_folder, and create the parent directory before copying.

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -858,6 +898,16 @@ def create_custom_env(self):
...
             custom_env['DOWNLOADED_PACKAGE_FOLDERS'] = ';'.join(package_folder_list)
+            print(f"  Adding env var DOWNLOADED_PACKAGE_FOLDERS={custom_env['DOWNLOADED_PACKAGE_FOLDERS']}")
+            path_env_var_sep = ';' if platform.system() == 'Windows' else ':'
+            custom_env['CMAKE_PREFIX_PATH']  = path_env_var_sep.join(package_folder_list)
+            print(f"  Adding env var CMAKE_PREFIX_PATH={custom_env['CMAKE_PREFIX_PATH']}")
+
+        custom_env_vars = self.package_info.set_env_vars
+        for key, value in custom_env_vars.items():
+            print(f"  Adding env var {key} = {value}")
+            custom_env[key] = value
```
EnterTheArcane, line 971: Could we preserve an existing CMAKE_PREFIX_PATH and protect the scripts internal environment variables? I tested this with an existing prefix, and it was replaced rather than extended. Because set_env_vars is applied last, it can also overwrite values such as TEMP_FOLDER and TARGET_INSTALL_ROOT, which later install code treats as trusted filesystem locations. It may be safer to append/prepend the dependency paths and either reject reserved names or apply the internal values last.

EnterTheArcane (changes requested):

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: ah, this slipped through accidentally.  I'll fix it

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: Hmm, yeah, I'll refactor this. I'll do 2 things.

* add a list of forbidden overrides such as TEMP_FOLDER / TARGET_INSTALL_ROOT used by the script itself.
* prepend rather than replace CMAKE_PREFIX_PATH, CMAKE_MODULE_PATH.  I still want to prepend, though, since we want to override any system searching and prefer any modules we supply.

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: done!  and I tested it with SQLITE and lua.

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: Done!  I verify that its temp or a subfolder of temp, and I verify that it creates the path

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: I've done that. Thanks for the suggestion!

Scripts/extras/pull_and_build_from_git.py
EnterTheArcane: The docs say mapped destinations are relative to `temp/src`, but this anchors them at `temp`. As a result, the documented `../dest/foo.txt` example is rejected as outside `temp`. Could we align the implementation and docs, and validate that mapped entries contain exactly two strings?

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -288,12 +308,18 @@ def _get_value(value_key, required=True, default=None):
...
+        self.clean_between_configurations = _get_value("clean_between_configurations", required=False, default=False)
+        self.set_env_vars = _get_value("set_env_vars", required=False, default={})
 
         if self.cmake_find_template and self.cmake_find_source:
             raise BuildError("Bad build config file. 'cmake_find_template' and 'cmake_find_source' cannot both be set in the configuration.")
         if not self.cmake_find_template and not self.cmake_find_source:
             raise BuildError("Bad build config file. 'cmake_find_template' or 'cmake_find_source' must be set in the configuration.")
 
+        for key, value in self.set_env_vars.items():
+            if key in FORBIDDEN_CUSTOM_ENV_VARS:
```
EnterTheArcane: Could we normalize reserved names and validate values here? On Windows, temp_folder bypasses the TEMP_FOLDER check because the comparison is case-sensitive. Numeric values are also accepted here but later cause subprocess.run() to raise a TypeError. Comparing uppercase names and requiring or converting values to strings should address both.

EnterTheArcane (changes requested):

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: good idea! I'm not too worried about intentionally feeding it bad data like
```json
{ 
   "213124" : "lol tricked you",
   "invalid-var" : "unix disallows dashes",
```
but I can at least do a quick check up front here for simple issues like case.

EnterTheArcane: Yeah, good point. I think those things would get caught pretty early. My brain tries to take a more defensive approach when possible. Just when I start to think there will never be untrustworthy data, it sneaks in...

Scripts/extras/pull_and_build_from_git.py
```diff
@@ -288,12 +308,28 @@ def _get_value(value_key, required=True, default=None):
...
         if not self.cmake_find_template and not self.cmake_find_source:
             raise BuildError("Bad build config file. 'cmake_find_template' or 'cmake_find_source' must be set in the configuration.")
 
+        for key, value in self.set_env_vars.items():
+            if type(value) is not str:
+                raise BuildError(f"Invalid build config file. 'set_env_vars' must be a dictionary of string key-value pairs but value '{value}' for key '{key}' is not a string.")
+            if key.lower() in FORBIDDEN_CUSTOM_ENV_VARS:
+                raise BuildError(f"Invalid build config file. 'set_env_vars' cannot set the reserved environment variable '{key}' to '{value}'.")
+
+        for element in self.additional_src_files:
```
EnterTheArcane: additional_src_files is optional and currently defaults to None, so this loop raises when the field is omitted. I reproduced it, and 33 of the package configs don't define this field. Could we default it to [] or guard this validation loop?

EnterTheArcane (changes requested):

Scripts/extras/pull_and_build_from_git.py
nick-l-o3de: done!

EnterTheArcane: Perfect! Great work!

EnterTheArcane (approved):

PR #400, opened 2026-08-15
Update OpenEXR to a newer version for prep of OpenImageIO

nick-l-o3de:
OpenImageIO requires a newer OpenEXR, and there's been a few security issues in OpenEXR anyway, so its about time to update.  

This depends on the IMath hash from https://github.com/o3de/3p-package-source/pull/394 - as it sits on "dev" bucket.  If we redo that one, we'll have to rev this one too.

This first attempt also produces packages for ios and android, which were not previously generated.  If this produces problems, I'll re-disable those, but openEXR is used for image dump and capture in the runtime, like, screenshot stuff,  (OPTIONALLY) and it would be nice to get those platforms up to spec for it.


Scripts/extras/pull_and_build_from_git.py
```diff
@@ -635,17 +635,27 @@ def sync_source(self):
...
-                        call_result = subprocess.run(subp_args(git_restore_cmd),
-                                                     shell=True,
-                                                     capture_output=False,
-                                                     cwd=str(self.src_folder.resolve()))
+                        git_clean_command = ['git', 'clean', '-fd']
+                        call_result = subprocess.run(subp_args(git_clean_command),
+                                                        shell=True,
+                                                        capture_output=False,
+                                                        cwd=str(self.src_folder.resolve()))
                         if call_result.returncode != 0:
```
nick-l-o3de, line 643: Note, this just adds a `clean -fd` before the `restore` command, since restore won't remove untracked files.

Scripts/cmake/o3de_package_utilities.cmake
```diff
@@ -129,13 +129,22 @@ macro(o3de_import_existing_config_files package_name search_dir)
     # for other platforms, we set the xxxx_ROOT value.
     set(_old_find_root_path ${CMAKE_FIND_ROOT_PATH})
     set(CMAKE_FIND_ROOT_PATH ${search_dir} ${CMAKE_FIND_ROOT_PATH})
+
+    set(_old_ignore_prefix_path ${CMAKE_IGNORE_PREFIX_PATH})
+    set(CMAKE_IGNORE_PREFIX_PATH ${CMAKE_CURRENT_LIST_DIR})
```
nick-l-o3de, line 134: Note, this prevents an infinite recursion problem, where this file would find package and its own folder is added to the package prefix paths.

spham-amzn (approved):

PR #403, opened 2026-08-22
Update Python to v3.14.7

EnterTheArcane:
## WHAT / WHY

- Updates Python to `3.14.7-rev1` on Windows-X64, Linux-X64, Linux-ARM, and Mac-ARM.
- Moves the Mac build onto `pull_and_build_from_git.py` and the O3DE `OpenSSL-3.6.3-rev1-mac-arm64` package instead of the installer's own OpenSSL download, and adds a `test-python.sh` validation step. Linux picks up `OpenSSL-3.6.3-rev1` in place of 1.1.1t.
- Adds a static zstd 1.5.7 to the Linux and Mac builds so the new `compression.zstd` stdlib module is present, and ships `LICENSE.ZSTD` with every package.
- Adds `Scripts/packaging/sign_macos_binaries.py`, which signs every Mach-O in the framework. Ad-hoc by default, or a Developer ID with hardened runtime and a secure timestamp when `3PS_MACOS_CODE_SIGN_IDENTITY` is set. Bytecode is precompiled for all three optimization levels before signing so first launch does not invalidate the sealed bundle.
- Updates the Mac installer patch for Tcl/Tk 9.0 and raises the deployment target to 13.0.
- Builds Windows with the VS2022 v143 toolset and drops the vendored libexpat copy and the `open3d_python.patch`, both of which upstream no longer needs.
- Replaces the `get-pip.py` download, manual pip/setuptools/wheel upgrades, and shebang rewriting with `--with-ensurepip=install` (`ensurepip --upgrade` on Windows).
- Fixes latent bugs in the Linux Docker build: `LIBFFI_GIT_URL` and `LIBFFI_LIB_PATH` were misspelled, several `retVal` checks tested a stale value, and the OpenSSL pkg-config files are now rewritten to the downloaded package prefix so a non-relocatable `.pc` does not break `configure`.
- Extends `quick_validate_python.py` to assert the exact version, reject a free-threaded build, import `compression.zstd`, `ensurepip`, `pip`, and `venv`, and print the SOABI.

`distutils` was removed in 3.12, so `--include-distutils` is gone from the Windows layout and consumers that still import it need setuptools. The standard GIL-enabled interpreter is used on every platform. The free-threaded build is intentionally not packaged.

## TEST

All GitHub Actions package builds and validations pass.


package-system/python/darwin_arm64/FindPython.cmake
nick-l-o3de: I recently learned a lesson about this one, and that is that sometimes, if we ever want to reuse "our python" when compiling "other libraries", for example if we want to use this python package when compiling, say, opencolorio's python modules, this early-exit actually breaks that.  I know its a pattern we use constantly, although, I no longer use it with my newer packages.

basically, I'd recommend changing this to not an early return, or only early return if the variable `Python_FOUND` evaluates to true, ie, `if(Python_FOUND)`.  Otherwise continue anyway and set all the various variables like "PYTHON_VERSION", "PYTHON_INTERPRETER_ID" and so on regardless of whether the target exists or not.  

 Then, only create/add the targets themselves, if the targets dont exist...

So more like
```cmake

set(....)
set(...)
set(...) # set all the globals like PackageName_VERSION etc and so on

if (NOT TARGET targetName)
    # add target and declare the target properties, etc.
endif()

set(libraryName_FOUND TRUE)
```

The subtle reason for this is that sometimes, instead of using targets, 3rd party libraries use those "Set variabels" like libraryName_INCLUDES or libraryName_INTERPRETER or whatever, and those are local variables scoped to directory not global.

An example in OpenImageIO is that it imports, say, ZLIB in one subfolder, like code/src, which creates the target globally since we declare zlib as `static imported GLOBAL`.  Then later on, one of its other 3p deps that it brings in also wants zlib, in a different subfolder, like externaldeps/whatever, but, because its a different folder, the **target** exists, but the variables do not.

